PR #4, opened 2026-04-30
Replace run-locally.sh with Docker Compose

tomi:
## Summary

- Replaces the ~130-line imperative `scripts/run-locally.sh` with declarative Docker Compose files (`compose.yaml`, `compose.macos.yaml`, `compose.linux.yaml`)
- The wrapper script shrinks to ~18 lines: detect arch, build images, select platform override, `docker compose up`
- Compose handles networking, health checks, dependency ordering, and container lifecycle natively
- A one-shot `sandbox-push` service tags and pushes the sandbox image to the local registry before runners start
- Adds `make up` / `make down` convenience targets

## Test plan

- [ ] `make up` brings up all services, API responds at `localhost:8080`
- [ ] `make up` with `-- -d` works for detached mode
- [ ] `make down` cleans everything up
- [ ] `curl -s -X POST http://localhost:8080/sandboxes -H 'X-Api-Key: test' | jq` creates a sandbox

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SHELL := /bin/bash
 MODULE  := github.com/n8n-io/sandbox-service
 BINDIR  := bin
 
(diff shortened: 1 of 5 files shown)


endor (approved):
LGTM

PR #14, opened 2026-05-06
chore: move CORS behind SANDBOX_API_ENABLE_CORS env var

tomi:
## Summary

- CORS was unconditionally enabled for all origins on the API gateway. This moves it behind a new `SANDBOX_API_ENABLE_CORS` env var (default `false`) so production doesn't expose CORS headers.
- Local dev compose sets the flag to `true` so the browser playground continues to work.
- Updated API.md and README.md with the new env var and corrected middleware list.

## Test plan

- [x] `make fmt-check` passes
- [x] `make vet` passes
- [x] `make test` passes (existing tests don't rely on CORS, default `false` is fine)
- [ ] `make up` + `make playground` — verify playground at localhost:5173 can still talk to the API

---
## Summary by cubic
Gated CORS behind the `SANDBOX_API_ENABLE_CORS` env var (default false) so the API no longer sends CORS headers by default in production.

- **Migration**
  - Set `SANDBOX_API_ENABLE_CORS=true` in environments that need browser access (e.g., the playground).
  - Leave unset/false for production to keep CORS disabled.

<sup>Written for commit b26fa19b6382922800b769bd04212805c4919327. Summary will update on new commits.</sup>

### DIFF
--- a/API.md
+++ b/API.md
@@ -9,7 +9,7 @@ Runtime topology:
 - If no runner is registered (or none are healthy / within capacity), operations that need a runner return **503** with a JSON error whose message explains that no runners are available.
 - Sandbox-scoped requests are proxied to the **stored** runner HTTP URL for that sandbox. If that runner is down or unreachable, the API returns **503** with JSON `error` **`runner unavailable`** (same shape as other API errors).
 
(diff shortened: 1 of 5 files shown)


endor (approved):
Cool. Was this based on https://www.notion.so/n8n/Sandbox-Security-Hardening-Plan-3585b6e0c94f8068a147f34995cc10d7?source=copy_link#3585b6e0c94f8021928fcbd9cdcdcc93?

tomi:
> Cool. Was this based on https://www.notion.so/n8n/Sandbox-Security-Hardening-Plan-3585b6e0c94f8068a147f34995cc10d7?source=copy_link#3585b6e0c94f8021928fcbd9cdcdcc93?

Yes. We only need the CORS for the playground, which is not required on prod

PR #17, opened 2026-05-07
chore: restart sandbox on kill

endor:
## Summary

* Make Docker restart containers that died
* Add e2e tests making sure that when a container dies it comes back up
* Add e2e tests verifying responses for OOMs
* Add basic documentation about failure modes

---
## Summary by cubic
Auto-restart sandboxes after exit/OOM so the same sandbox ID keeps working on the same runner. While restarting or without network we return 503 (“sandbox unavailable”); if the inner container isn’t running we return 502 (“not retryable”) so clients change strategy.

- **New Features**
  - Start sandboxes with Docker restart policy `unless-stopped`.
  - Clear 503 vs 502: 503 for transient runner/daemon/network unavailability; 502 when the sandbox exists but isn’t running. Readiness requires `/healthz` plus a tiny `/exec`.
  - Docs: failure modes in README; API.md explains 503 (retry) vs 502 (don’t retry).
  - e2e: restart keeps same ID, 503 while restarting/network-not-ready, and 502 while container is stopped (and recovery after start).
  - `sdk`: configurable retries via `SandboxClientOptions.retry`; defaults retry 429/503 with backoff/jitter; non-idempotent calls can opt in per request with `isSafeToRetry` (502 is not retried by default).

- **Bug Fixes**
  - Structured exec errors when a command is SIGKILLed or the daemon dies mid-exec; covered by new tests.
  - Network rule teardown no longer needs a container IP; delete/reconcile works even if the container isn’t running or has no network.

<sup>Written for commit 35b0760816ed8005508bcbf9a77eedbb7cd4d479. Summary will update on new commits.</sup>

### DIFF
--- a/e2e/tests/network-isolation.spec.ts
+++ b/e2e/tests/network-isolation.spec.ts
@@ -11,6 +11,41 @@ const pyConnect = (ip: string, port: number = 80, timeout: number = 3) =>
 const pyResolve = (host: string) =>
   `python3 -c "import socket; print(socket.gethostbyname('${host}'))"`;
 
+function isTransientExecError(err: unknown): boolean {
+  const status = (err as { status?: number })?.status;
+  if (status === 502 || status === 503) {
+    return true;
+  }
+  const msg = String((err as Error)?.message || '').toLowerCase();
+  return (
+    msg.includes('internal server error') ||
+    msg.includes('daemon unreachable') ||
+    msg.includes('sandbox exec stream ended without an exit event')
+  );
+}
+
+async function execWithTransientRetry(
+  id: string,
+  command: string,
+  timeoutMs: number,
+  retryWindowMs: number = 12_000,
+) {
+  const deadline = Date.now() + retryWindowMs;
+  let lastErr: unknown;
+  while (Date.now() < deadline) {
+    try {
+      return await exec(id, command, { timeoutMs });
+    } catch (err) {
+      lastErr = err;
+      if (!isTransientExecError(err)) {
+        throw err;
+      }
+      await new Promise((r) => setTimeout(r, 200));
+    }
+  }
+  throw new Error(`exec did not recover within ${retryWindowMs}ms: ${String(lastErr)}`);
+}
+
 test.describe('Network isolation', () => {
   test('sandbox can reach public internet', async () => {
     const id = await createSandbox();
@@ -36,9 +71,7 @@ test.describe('Network isolation', () => {
       ];
 
       for (const ip of privateIPs) {
-        const result = await exec(id, pyConnect(ip, 80, 3), {
-          timeoutMs: 10_000,
-        });
+        const result = await execWithTransientRetry(id, pyConnect(ip, 80, 3), 10_000);
         expect(result.exitCode, `expected private IP ${ip} to be unreachable`).not.toBe(0);
       }
     } finally {
@@ -51,25 +84,23 @@ test.describe('Network isolation', () => {
     const id2 = await createSandbox();
     try {
       // Get sandbox 2's IP
-      const ipResult = await exec(id2, 'hostname -I', { timeoutMs: 5_000 });
+      const ipResult = await execWithTransientRetry(id2, 'hostname -I', 5_000);
       expect(ipResult.exitCode).toBe(0);
       const sandbox2IP = ipResult.stdout.trim();
       expect(sandbox2IP).toBeTruthy();
 
       // Start a TCP listener in sandbox 2 on port 9999
-      await exec(id2, `python3 -c "
+      await execWithTransientRetry(id2, `python3 -c "
 import socket, threading
 s = socket.socket()
 s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 s.bind(('0.0.0.0', 9999))
 s.listen(1)
 import time; time.sleep(30)
-" &`, { timeoutMs: 5_000 });
+" &`, 5_000);
 
       // Sandbox 1 should not be able to reach sandbox 2
-      const result = await exec(id1, pyConnect(sandbox2IP, 9999, 3), {
-        timeoutMs: 10_000,
-      });
+      const result = await execWithTransientRetry(id1, pyConnect(sandbox2IP, 9999, 3), 10_000);
       expect(result.exitCode, `expected sandbox 1 to not reach sandbox 2 at ${sandbox2IP}`).not.toBe(0);
     } finally {
       await deleteSandbox(id1);
@@ -80,7 +111,7 @@ import time; time.sleep(30)
   test('DNS resolution works', async () => {
     const id = await createSandbox();
     try {
-      const result = await exec(id, pyResolve('example.com'), { timeoutMs: 10_000 });
+      const result = await execWithTransientRetry(id, pyResolve('example.com'), 10_000);
       expect(result.exitCode).toBe(0);
       // Should resolve to an IP address
       expect(result.stdout.trim()).toMatch(/^\d+\.\d+\.\d+\.\d+$/);
--- a/e2e/tests/network-isolation.spec.ts
+++ b/e2e/tests/network-isolation.spec.ts
@@ -90,14 +90,14 @@ test.describe('Network isolation', () => {
       expect(sandbox2IP).toBeTruthy();
 
       // Start a TCP listener in sandbox 2 on port 9999
-      await execWithTransientRetry(id2, `python3 -c "
+      await exec(id2, `python3 -c "
 import socket, threading
 s = socket.socket()
 s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 s.bind(('0.0.0.0', 9999))
 s.listen(1)
 import time; time.sleep(30)
-" &`, 5_000);
+" &`, { timeoutMs: 5_000 });
 
       // Sandbox 1 should not be able to reach sandbox 2
       const result = await execWithTransientRetry(id1, pyConnect(sandbox2IP, 9999, 3), 10_000);
--- a/e2e/tests/network-isolation.spec.ts
+++ b/e2e/tests/network-isolation.spec.ts
@@ -1,5 +1,5 @@
 import { test, expect } from '@playwright/test';
-import { createSandbox, deleteSandbox, exec } from './helpers';
+import { createSandbox, deleteSandbox, exec, execWithTransientRetry } from './helpers';
 
 // Helper: use python3 for HTTP requests since curl is not in the sandbox rootfs
 const pyGet = (url: string, timeout: number = 5) =>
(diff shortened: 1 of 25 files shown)


e2e/tests/sandbox-recovery.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    test.setTimeout(150_000);
+
+    const runnerContainer = process.env.E2E_RUNNER_CONTAINER_NAME!;
+    const id = await createSandbox();
+    const innerName = innerContainerName(id);
+    try {
+      const before = inspectRestartState(runnerContainer, innerName);
+      // Force a full container restart from runner-side Docker and verify the
+      // sandbox ID remains usable afterwards.
+      docker([
```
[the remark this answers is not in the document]
endor: We haven't been able to reliably produce a container restart due to OOM, so we'll have to live with this.

internal/runner/manager/manager.go
```diff
@@ -190,16 +193,30 @@ func (m *Manager) DaemonURL(ctx context.Context, sandboxID string) (string, erro
...
+	inspect, err := m.inspectContainer(ctx, containerID)
 	if err != nil {
+		if isDockerNotFound(err) {
+			return "", ErrSandboxNotFound
+		}
 		return "", err
 	}
-	return fmt.Sprintf("http://%s:%d", info.IP, daemonPort), nil
+	if !inspect.State.Running {
+		return "", ErrSandboxNotFound
```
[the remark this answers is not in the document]
endor: Fixed.

sdk/src/http.ts
```diff
@@ -22,10 +32,17 @@ export class HttpClient {
...
+      baseDelayMs: Math.max(0, retry?.baseDelayMs ?? 200),
+      maxDelayMs: Math.max(0, retry?.maxDelayMs ?? 2000),
+      retryOnStatuses: new Set(retry?.retryOnStatuses ?? [429, 502, 503]),
+      jitter: retry?.jitter ?? true,
+    };
   }
 
   async requestJson<T>(method: Method, path: string, options: RequestOptions = {}): Promise<T> {
-    try {
+    return this.withRetry(async () => {
```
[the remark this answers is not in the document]
endor: Fixed.

e2e/tests/network-isolation.spec.ts
```diff
@@ -51,25 +84,23 @@ test.describe('Network isolation', () => {
...
-      await exec(id2, `python3 -c "
+      await execWithTransientRetry(id2, `python3 -c "
 import socket, threading
 s = socket.socket()
 s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 s.bind(('0.0.0.0', 9999))
 s.listen(1)
 import time; time.sleep(30)
-" &`, { timeoutMs: 5_000 });
+" &`, 5_000);
```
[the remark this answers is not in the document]
endor: Fixed.

sdk/src/types.ts
```diff
@@ -4,6 +4,28 @@ export interface SandboxClientOptions {
...
+/** Retry policy for transient HTTP failures. */
+export interface RetryOptions {
+  /**
+   * Number of retry attempts after the initial request.
+   * Example: 3 means at most 4 total attempts.
+   */
+  attempts?: number;
+  /** Initial backoff delay in milliseconds. Defaults to 200. */
+  baseDelayMs?: number;
+  /** Maximum backoff delay in milliseconds. Defaults to 2000. */
```
tomi: 2s can be quite short, maybe something like 5-10s would be a decent default?

sdk/src/types.ts
```diff
@@ -4,6 +4,28 @@ export interface SandboxClientOptions {
   apiKey?: string;
   /** Base URL of the sandbox service API. */
   baseUrl?: string;
+  /** Optional retry policy for transient request failures. */
+  retry?: RetryOptions;
+}
+
+/** Retry policy for transient HTTP failures. */
+export interface RetryOptions {
```
tomi, line 15: 👍 for adding these config options. I also added retrying on my PR. We should align so that the other PR aligns with the one that gets merged first

sdk/src/http.ts
```diff
@@ -1,18 +1,30 @@
...
 import { SandboxServiceError, createErrorFromResponse } from "./errors";
+import type { RetryOptions } from "./types";
 
 export interface RequestOptions {
   data?: unknown;
   params?: Record<string, string>;
   headers?: Record<string, string>;
   signal?: AbortSignal;
+  /** Allow retry on non-idempotent methods (e.g. POST). Default false. */
+  retryUnsafe?: boolean;
```
tomi: This name reads a bit wonky. How about `isSafeToRetry`?

sdk/src/http.ts
```diff
@@ -35,17 +54,15 @@ export class HttpClient {
...
+    }, method, options.signal, options.retryUnsafe === true);
   }
 
   async requestStream(
     method: Method,
     path: string,
     options: RequestOptions = {},
   ): Promise<{ stream: Readable; status: number }> {
-    try {
+    return this.withRetry(async () => {
```
tomi: I don't think we can apply plain retrying on streaming requests by default. Being able to retry the stream requires support from the target endpoint to be able to figure out from which point to retry. If we retry without this knowledge it would restart the stream from beginning, which might work on some cases but might cause issues on other cases. E.g. if we have a per stream line handler attached we would cause duplicate events/callbacks if it restarts mid stream. So IMHO the default for stream should be false

internal/runner/handlers.go
```diff
@@ -129,18 +129,7 @@ func DeleteSandbox(mgr ContainerManager) http.HandlerFunc {
...
-			if errors.Is(err, manager.ErrSandboxNotFound) {
-				writeError(w, http.StatusNotFound, err.Error())
-			} else {
-				writeError(w, http.StatusInternalServerError, err.Error())
-			}
-			return
-		}
-
-		if err := mgr.DeleteContainer(r.Context(), containerID, containerInfo.IP); err != nil {
+		if err := mgr.DeleteContainer(r.Context(), containerID); err != nil {
```
tomi, line 132: nice that we don't need this part 👍

internal/runner/manager/manager.go
```diff
@@ -118,11 +128,7 @@ func (m *Manager) CreateContainer(ctx context.Context, sandboxID string, opts *C
 	}
 
 	cleanupOnError := func(containerIP string) {
```
tomi, line 130: The containerIP doesn't seem to be used here anymore?

internal/runner/manager/manager.go
```diff
@@ -255,7 +275,33 @@ func waitForDaemon(ctx context.Context, baseURL string) error {
 			}
 			io.Copy(io.Discard, resp.Body)
 			resp.Body.Close()
-			if resp.StatusCode == http.StatusOK {
+			if resp.StatusCode != http.StatusOK {
+				continue
+			}
+
+			// /healthz can become ready slightly before command execution is fully
+			// usable under load; require a tiny exec round-trip before returning.
```
tomi, line 283: Any idea what's the reason for this? Could we make the healthcheck more robust instead of needing to rely on the exec? Or optionally don't do /healthz check at all and solely rely on this

internal/runner/manager/manager.go
```diff
@@ -255,7 +275,33 @@ func waitForDaemon(ctx context.Context, baseURL string) error {
...
+			if err != nil {
+				continue
+			}
+			body, _ := io.ReadAll(execResp.Body)
+			execResp.Body.Close()
+			if execResp.StatusCode != http.StatusOK {
+				continue
+			}
+			// Daemon /exec streams NDJSON events; require a successful exit event.
+			if bytes.Contains(body, []byte(`"type":"exit"`)) && bytes.Contains(body, []byte(`"exit_code":0`)) {
```
tomi: Can we move this part to a separate method?

internal/runner/manager/manager.go
```diff
@@ -268,12 +314,12 @@ func (m *Manager) reconcileContainers(ctx context.Context) error {
 		return err
 	}
 	for _, id := range ids {
-		inspect, inspectErr := m.inspectContainer(ctx, id)
+		_, inspectErr := m.inspectContainer(ctx, id)
 		if inspectErr == nil {
```
tomi, line 322: I see a slight risk here. If we first teardown the network rules, the container has a window where its network access is not limited, but it hasn't been removed yet. If the removal fails for whatever reason, it can be left in unsafe state. I'll create a separate issue for this.

e2e/tests/sandbox-recovery.spec.ts
```diff
@@ -0,0 +1,205 @@
+import { expect, test } from '@playwright/test';
+import { execFileSync } from 'node:child_process';
+import { createSandbox, deleteSandbox, exec } from './helpers';
+
+function docker(args: string[]): string {
```
tomi: Can we move the docker helpers into a separate file?

e2e/tests/sandbox-recovery.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  test('client sees 503 when sandbox is permanently stopped', async () => {
+    test.skip(!process.env.E2E_RUNNER_CONTAINER_NAME, 'needs E2E_RUNNER_CONTAINER_NAME (from e2e/run.sh)');
+    test.setTimeout(60_000);
+
+    const runnerContainer = process.env.E2E_RUNNER_CONTAINER_NAME!;
+    const id = await createSandbox();
+    const innerName = innerContainerName(id);
+    try {
+      docker(['exec', runnerContainer, 'docker', 'update', '--restart', 'no', innerName]);
```
tomi: Do we need this? I think the policy is "restart if not stopped", so it shouldn't restart if manually stopped

sdk/test/http.test.ts
```diff
@@ -75,4 +75,181 @@ describe("HttpClient", () => {
       });
     }
   });
+
+  it("retries transient 503 responses and eventually succeeds", async () => {
+    let hits = 0;
+    const retryServer = createServer((req, res) => {
```
tomi: Can we extract this test server creation to a separate function since it's used in many of these tests?

tomi (commented):
Nice stuff 👏 Especially good test coverage 👌 Few comments

sdk/src/http.ts
endor: The default is `false`, isn't it? Unless one explicitly provides `{isSafeToRetry: true}` in the options, `options.isSafeToRetry === true` is not the case and therefore we won't retry.

sdk/src/types.ts
endor: Fixed.

sdk/src/http.ts
endor: Fixed.

internal/runner/manager/manager.go
endor: Fixed.

e2e/tests/sandbox-recovery.spec.ts
endor: Fixed.

e2e/tests/sandbox-recovery.spec.ts
endor: Fixed.

sdk/test/http.test.ts
endor: Fixed.

tomi (approved):
🚀

PR #20, opened 2026-05-07
feat: add resumable exec sessions with disconnect survival

tomi:
## Summary

- Exec commands now run in daemon-side sessions whose lifetime is independent of the HTTP connection — disconnecting no longer kills the process.
- Each session is identified by an `exec_id` (client-provided or server-generated) and streams events with monotonically increasing `seq` numbers, enabling clients to resume from the last received event.
- New `GET /sandboxes/{id}/exec/{exec_id}` endpoint to resume/replay a session stream (`?after=<seq>&follow=true`), and `DELETE /sandboxes/{id}/exec/{exec_id}` to cancel a running command.
- The SDK transparently retries transient errors (connection resets, timeouts) with automatic resume via a new `ExecStreamConsumer` class, up to 10 retries.
- Sessions are retained for 10 minutes after completion with a 16 MiB bounded event buffer per session, configurable via `SANDBOX_EXEC_MAX_EVENT_BYTES` and `SANDBOX_EXEC_SESSION_RETAIN` env vars.

## Test plan

- [ ] Verify new `session_test.go` unit tests pass for session lifecycle, GC, history trimming, and concurrent followers
- [ ] Verify updated `sdk/test/exec.test.ts` tests pass for resume, cancel, and transient error retry logic
- [ ] Run `e2e/tests/resumable-exec.spec.ts` end-to-end tests
- [ ] Manually test disconnect/reconnect scenario against a running sandbox

---
## Summary by cubic
Adds resumable exec executions so commands keep running if the HTTP stream drops. Clients can reconnect from the last event; the `@n8n/sandbox-client` SDK auto-resumes (including on truncated streams via typed errors) with up to 10 retries, and aborts now reliably cancel and delete the execution.

- **New Features**
  - Daemon-side executions identified by `exec_id`; first event is `{"seq":0,"type":"started","exec_id":...}`, all events include monotonic `seq`.
  - POST `/sandboxes/{id}/exec` accepts optional `exec_id` (idempotent start). GET `/sandboxes/{id}/exec/{exec_id}?after=<seq>&follow=true` replays/follows. DELETE `/sandboxes/{id}/exec/{exec_id}` cancels and deletes the execution; idempotent (`204` even if missing). GET returns `404` when missing and `410` when requested history was trimmed.
  - Executions survive disconnects with history buffered up to 16 MiB and retained 10 minutes (config via `SANDBOX_EXEC_MAX_EVENT_BYTES`, `SANDBOX_EXEC_RETAIN`; see README). Streaming is faster via O(1) event indexing.
  - `@n8n/sandbox-client` auto-resumes on transient errors and truncated NDJSON streams (throws `InvalidStreamEventError`) using `ExecStreamConsumer` with up to 10 retries. Abort signals trigger an automatic DELETE using the client-defined `exec_id`. New helpers: `resumeExecution(sandboxId, execId, afterSeq)` and `deleteExecution(sandboxId, execId)`.

- **Bug Fixes**
  - Fixed overflow when `after == MaxUint64` in resume indexing, avoiding unintended replays or incorrect history checks.

<sup>Written for commit d97d3583760d828954fa7db6ee2d629bae2212de. Summary will update on new commits.</sup>


endor:
Why not go for `session` instead of `exec`? Seems like a more easily understood concept. (`exec_id` -> `session_id`, `GET /sandboxes/{id}/exec/{exec_id}` -> `GET /sandboxes/{id}/sessions/{session_id}`, etc.)

tomi:
> Why not go for `session` instead of `exec`? Seems like a more easily understood concept. (`exec_id` -> `session_id`, `GET /sandboxes/{id}/exec/{exec_id}` -> `GET /sandboxes/{id}/sessions/{session_id}`, etc.)

Session has a very specific meaning in unix and shells, where a session is shared among multiple process executions. What we are doing here is single process execution. I wanted to keep the concepts now separate, as we might want to introduce session support later on, where you could create a session and execute multiple processes within that session

internal/api/gateway.go
```diff
@@ -26,6 +26,8 @@ func NewGatewayRouter(s *store.Store, cfg *config.APIConfig, reg *registry.Regis
 	mux.HandleFunc("DELETE /sandboxes/{id}", handleDeleteSandbox(s, cfg))
 
 	mux.HandleFunc("POST /sandboxes/{id}/exec", sandboxProxy(false))
+	mux.HandleFunc("GET /sandboxes/{id}/exec/{exec_id}", sandboxProxy(false))
```
endor, line 29: I know this is nitpicky, but could we use `executions` or `execution_sessions`? Feels more REST-y.

tomi: Agreed! I will do that as a followup, as this PR is already quite large

internal/daemon/execution.go
```diff
@@ -0,0 +1,154 @@
...
+	ex.notify = make(chan struct{})
+	ex.mu.Unlock()
+	close(old)
+}
+
+// eventsAfter returns pre-marshaled data for retained events with seq > after.
+// If after is nil, all retained events are returned. Caller must hold ex.mu.
+func (ex *Execution) eventsAfter(after *uint64) [][]byte {
+	var result [][]byte
+	for _, e := range ex.events {
```
endor: GPT 5.5 came up with an interesting observation: if we have a stream of a lot of events that arrive over time, then this function would loop through these many events many times. If we on the other hand passed an index indicating where we left off or did a binary search, we could make this more performant.

tomi: Hmmm I don't know if that is true 🤔  We take a lock on each case where this is called, so there shouldn't be two code paths mutating the events slice

tomi:
@cubic-dev-ai review

internal/api/middleware.go
```diff
@@ -99,12 +103,52 @@ func CORSMiddleware(next http.Handler) http.Handler {
...
+
+func (rw *recoveryResponseWriter) WriteHeader(code int) {
+	if rw.wroteHeader {
+		return
+	}
+	rw.wroteHeader = true
+	rw.ResponseWriter.WriteHeader(code)
+}
+
+func (rw *recoveryResponseWriter) Write(data []byte) (int, error) {
```
endor, line 136: Could make sense to document these functions. What are they used for?

endor (approved):

internal/api/middleware.go
tomi: I'll do as a followup

tomi: https://github.com/n8n-io/n8n-sandbox-service/pull/21

PR #32, opened 2026-05-12
feat: add sandbox container lifecycle management

endor:
**Summary**

* Added configurable `stopped` TTL, after which the sandbox is stopped
* Added configurable `deleted` TTL, after which the sandbox is deleted
* Removed deprecated `terminated` sandbox state

---
## Summary by cubic
Adds idle lifecycle management for sandboxes with configurable stop/delete TTLs and a background sweeper. Stopped sandboxes auto‑wake on demand; past the delete window, the API returns 404 as if the sandbox never existed. Meets Linear CP-2639.

- **New Features**
  - Configurable idle TTLs with background sweeper: `SANDBOX_API_IDLE_STOP_AFTER`, `SANDBOX_API_IDLE_DELETE_AFTER`, `SANDBOX_API_IDLE_DELETE_SAFETY_BUFFER`, `SANDBOX_API_IDLE_SWEEP_INTERVAL`.
  - API: returns 404 for sandboxes past delete-after (GET and proxy); updates `last_active_at` on GET and on proxied request receipt; status becomes `running` on proxy receipt and `stopped` after idle stop.
  - Runner: added gRPC `StopSandbox`; proxy auto‑wakes stopped containers via `EnsureSandboxRunning` with de‑duped wake using `golang.org/x/sync`; waits for the daemon before routing.
  - E2E: new idle‑TTL stack (`e2e/run-idle-ttl.sh`) and spec; `run-all.sh` adds an idle‑TTL phase; full suite gains transient create retry; recovery tests updated for auto‑wake.

- **Migration**
  - Idle sweeper is on by default (`stop_after=1h`, `delete_after=24h`, buffer `1m`). Set both to `0` to disable.
  - Clients may now see 404 for sandboxes past the delete window; stopped sandboxes are auto‑woken instead of returning 502.
  - The deprecated `terminated` state is removed; use `running`/`stopped`.
  - Deploy the updated runner to provide `StopSandbox` and auto‑wake.

<sup>Written for commit 4f8adfd5e638a42c9569853cc8c6cabcb3ecc0a5. Summary will update on new commits. <a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/32?utm_source=github">Review in cubic</a></sup>


internal/api/config/config.go
```diff
@@ -161,5 +173,39 @@ func LoadAPI() (*APIConfig, error) {
...
+		cfg.IdleDeleteAfter = d
+	}
+	if v := strings.TrimSpace(os.Getenv("SANDBOX_API_IDLE_DELETE_SAFETY_BUFFER")); v != "" {
+		d, err := time.ParseDuration(v)
+		if err != nil || d < 0 {
+			return nil, fmt.Errorf("SANDBOX_API_IDLE_DELETE_SAFETY_BUFFER must be a non-negative duration, got %q", v)
+		}
+		cfg.IdleDeleteSafetyBuffer = d
+	}
+	if cfg.IdleDeleteAfter > 0 && cfg.IdleDeleteSafetyBuffer <= 0 {
```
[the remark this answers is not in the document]
endor: This is acceptable, because we always need this to be set to something other than 0.

internal/api/ttl.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	go func() {
+		t := time.NewTicker(interval)
+		defer t.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-t.C:
+				sweepIdleSandboxes(context.Background(), s, cfg, tlsCfg)
```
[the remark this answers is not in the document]
endor: Fixed.

internal/runner/manager/manager.go
```diff
@@ -190,6 +191,67 @@ func (m *Manager) GetContainerInfo(ctx context.Context, containerID string) (*Co
 	}, nil
 }
 
+// EnsureSandboxRunning starts a stopped container if needed, reapplies network
+// policy for its current IP, and waits until the daemon accepts traffic.
+func (m *Manager) EnsureSandboxRunning(ctx context.Context, sandboxID string) error {
+	_, err, _ := m.wakeGroup.Do(sandboxID, func() (interface{}, error) {
+		return nil, m.ensureSandboxRunningOnce(ctx, sandboxID)
+	})
```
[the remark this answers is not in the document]
endor: Fixed.

internal/api/config/config.go
```diff
@@ -161,5 +173,39 @@ func LoadAPI() (*APIConfig, error) {
 		return nil, fmt.Errorf("SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_CA_FILE, SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_CERT_FILE, and SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_KEY_FILE are required for control-plane mTLS")
 	}
 
+	if v := strings.TrimSpace(os.Getenv("SANDBOX_API_IDLE_STOP_AFTER")); v != "" {
+		d, err := time.ParseDuration(v)
+		if err != nil || d <= 0 {
```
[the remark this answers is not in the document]
endor: Fixed.

README.md
```diff
@@ -110,6 +110,12 @@ curl -s -X DELETE http://localhost:8080/sandboxes/<id> \
 | `SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_CERT_FILE` | *(required)* | API client certificate (PEM) for SandboxControl |
 | `SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_KEY_FILE` | *(required)* | API client key (PEM) |
 | `SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_SERVER_NAME` | *(empty)* | TLS verify name when it must differ from the dial host (defaults to the runner host) |
+| `SANDBOX_API_IDLE_STOP_AFTER` | *(unset / off)* | After this duration since `last_active_at`, the API calls runner **StopSandbox** (container kept). Go [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) (e.g. `30m`, `2h`). `0` means the same as unset (idle stop disabled). |
+| `SANDBOX_API_IDLE_DELETE_AFTER` | *(unset / off)* | After this duration since `last_active_at`, proxied routes and `GET /sandboxes/{id}` return **404** (same as missing sandbox) until the sweeper removes the runner container and DB row after the safety buffer. `0` means the same as unset (idle delete disabled). |
```
tomi: IMO we should have some default for these. E.g. 1h for stop, 7d for delete or something

internal/runner/manager/manager.go
```diff
@@ -190,6 +191,75 @@ func (m *Manager) GetContainerInfo(ctx context.Context, containerID string) (*Co
...
+		return nil
+	}
+	if err := m.startContainer(ctx, containerID); err != nil {
+		return fmt.Errorf("start container: %w", err)
+	}
+	containerIP, err := m.containerIP(ctx, containerID)
+	if err != nil {
+		return err
+	}
+	if err := netrules.ApplyPolicy(containerID, containerIP, m.gatewayIP, daemonPort); err != nil {
```
tomi: I think there might a risk of a timing attack here, since the network policy is applied only after the container has started. Or we do forbid somewhere traffic to the container before we have finished initialization?

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+		deleteSec = int64(cfg.IdleDeleteAfter.Seconds())
+		bufferSec = int64(cfg.IdleDeleteSafetyBuffer.Seconds())
+		deleteCutoff = now - deleteSec - bufferSec
+	}
+	if cfg.IdleStopAfter > 0 {
+		stopSec = int64(cfg.IdleStopAfter.Seconds())
+		stopCutoff = now - stopSec
+	}
+
+	if cfg.IdleDeleteAfter > 0 {
```
tomi: nit: handling of delete and stop could be their own functions

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+		records, err := s.ListForIdleReapDelete(deleteCutoff)
+		if err != nil {
+			slog.Info("idle sweep list delete candidates failed", "err", err)
```
tomi: `.Info` or `.Error`?

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+		} else {
+			for _, rec := range records {
+				if rec == nil {
+					continue
+				}
+				if err := runnerctl.DeleteSandbox(ctx, rec.RunnerControlGRPCAddr, cfg.RunnerAPIKey, tlsCfg, rec.ID); err != nil {
+					if ctx.Err() != nil {
```
tomi: What is this case for? For cancellation?

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+						return
+					}
+					slog.Info("idle delete failed", "sandbox_id", rec.ID, "err", err)
```
tomi: `.Error`?

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+					continue
+				}
+				if err := s.Delete(rec.ID); err != nil {
+					slog.Info("idle delete store failed", "sandbox_id", rec.ID, "err", err)
```
tomi: `.Error`?

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+				}
+			}
+		}
+	}
+
+	if ctx.Err() != nil {
```
tomi: Same Q here? Is this for cancellation? A comment could clarify it

internal/api/ttl.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+
+	if cfg.IdleStopAfter > 0 {
+		records, err := s.ListForIdleReapStop(stopCutoff)
+		if err != nil {
+			slog.Info("idle sweep list stop candidates failed", "err", err)
```
tomi: And same comment here and below as well. Do we want to log errors here with info or warn or error?

internal/api/handlers.go
```diff
@@ -59,8 +93,20 @@ func sandboxProxyHandler(s *store.Store, cfg *config.APIConfig) func(bool) http.
 				return
 			}
 
+			if cfg.IdleDeleteAfter > 0 {
+				now := time.Now().Unix()
+				if now > rec.LastActiveAt+int64(cfg.IdleDeleteAfter.Seconds()) {
```
tomi: nit: Could be its own fn `isPastDeleteTtl` or similar

internal/api/handlers.go
```diff
@@ -114,6 +160,13 @@ func handleGetSandbox(s *store.Store) http.HandlerFunc {
 			writeError(w, http.StatusNotFound, "sandbox not found")
 			return
 		}
+		if cfg.IdleDeleteAfter > 0 {
```
tomi: Same check as above, could unify

internal/api/handlers.go
```diff
@@ -59,8 +93,20 @@ func sandboxProxyHandler(s *store.Store, cfg *config.APIConfig) func(bool) http.
...
+					writeError(w, http.StatusNotFound, "sandbox not found")
+					return
+				}
+			}
+
 			u, _ := url.Parse(strings.TrimRight(rec.RunnerHTTPBase, "/"))
-			proxy := newRunnerReverseProxy(u, cfg.RunnerAPIKey, cfg)
+			workDone := func() {
+				_ = s.UpdateLastActive(id)
+				_ = s.UpdateStatus(id, "running")
```
tomi: It feels a bit wrong to update the sandbox status here. But this is a bit bigger topic so let's discuss it separately

internal/api/handlers.go
```diff
@@ -59,8 +93,20 @@ func sandboxProxyHandler(s *store.Store, cfg *config.APIConfig) func(bool) http.
...
+				now := time.Now().Unix()
+				if now > rec.LastActiveAt+int64(cfg.IdleDeleteAfter.Seconds()) {
+					writeError(w, http.StatusNotFound, "sandbox not found")
+					return
+				}
+			}
+
 			u, _ := url.Parse(strings.TrimRight(rec.RunnerHTTPBase, "/"))
-			proxy := newRunnerReverseProxy(u, cfg.RunnerAPIKey, cfg)
+			workDone := func() {
```
tomi: I think it would be a good enough simplification to update the last active timestamp immediately and not after the request has finished. Yes it makes the TTL a bit off, but I think it's still good enough and would avoid the complexity in tracking the request finishing

e2e/tests/helpers.ts
```diff
@@ -20,6 +20,38 @@ export async function createSandbox(): Promise<string> {
...
+  return (
+    msg.includes('timeout waiting for daemon') ||
+    msg.includes('connect to daemon') ||
+    msg.includes('daemon temporarily unavailable') ||
+    msg.includes('runner unavailable') ||
+    msg.includes('internal server error')
+  );
+}
+
+export async function createSandboxWithRetry(maxAttempts = 5): Promise<string> {
```
tomi, line 38: Should this retrying be in the SDK?

tomi (commented):
Nice 👌 Some comments

internal/runner/manager/manager.go
endor: My understanding: clients aren't yet aware of the sandbox ID at this point, so no user code can run on the sandbox at this point. The only way to exploit this is if the sandbox image itself is compromised.

internal/api/ttl.go
endor: This is for a clean shutdown of the API without misleading errors being logged about deleting not being successful.

e2e/tests/helpers.ts
endor: Maybe. Sandbox creation is not idempotent, so if creation succeeds, but failure is reported back, we will have (a lot of) orphans.

internal/api/handlers.go
endor: Fixed.

internal/api/handlers.go
endor: Left there for now..

internal/api/handlers.go
endor: Fixed.

internal/api/handlers.go
endor: Fixed.

internal/api/ttl.go
endor: Fixed.

internal/api/ttl.go
endor: Fixed.

internal/api/ttl.go
endor: Fixed.

internal/api/ttl.go
endor: Fixed.

internal/api/ttl.go
endor: Fixed.

internal/api/ttl.go
endor: Fixed.

README.md
endor: Fixed.

internal/api/ttl.go
```diff
@@ -0,0 +1,121 @@
...
+				// Use the same ctx as the ticker loop so stop/delete RPCs honor
+				// sweepCancel() from shutdown instead of blocking on Background.
+				now := time.Now().Unix()
+				if cfg.IdleDeleteAfter > 0 {
+					sweepIdleDeleteSandboxes(ctx, s, cfg, tlsCfg, now)
+				}
+				if ctx.Err() != nil {
+					return
+				}
+				if cfg.IdleStopAfter > 0 {
```
tomi, line 60: If someone configures delete after == stop after, this will keep the stopped sandboxes until next sweep round. Is there a reason for deleting before stopping?

internal/api/config/config.go
```diff
@@ -49,6 +51,18 @@ type APIConfig struct {
 	GRPCServerKeyFile  string
 	GRPCClientCAFile   string
 
+	// IdleStopAfter is how long after last activity the API asks the runner to stop
+	// the container (0 = disabled).
+	IdleStopAfter time.Duration
+	// IdleDeleteAfter is how long after last activity the API deletes the sandbox
+	// (0 = disabled). Wakes are refused after this window until the row is removed.
+	IdleDeleteAfter time.Duration
+	// IdleDeleteSafetyBuffer is added to IdleDeleteAfter before deletion (race guard).
```
tomi, line 60: Race guard against what?

internal/runner/manager/manager.go
```diff
@@ -157,6 +158,74 @@ func (m *Manager) GetContainerInfo(ctx context.Context, containerID string) (*Co
 	}, nil
 }
 
+// EnsureSandboxRunning starts a stopped container if needed, reapplies network
+// policy for its current IP, and waits until the daemon accepts traffic.
+func (m *Manager) EnsureSandboxRunning(ctx context.Context, sandboxID string) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	_, err, _ := m.wakeGroup.Do(sandboxID, func() (interface{}, error) {
```
tomi, line 167: 👍 for handling multi in-flights

internal/runner/manager/manager.go
```diff
@@ -157,6 +158,74 @@ func (m *Manager) GetContainerInfo(ctx context.Context, containerID string) (*Co
...
+		return err
+	}
+	inspect, err := m.docker.inspectContainer(ctx, containerID)
+	if err != nil {
+		if isDockerNotFound(err) {
+			return ErrSandboxNotFound
+		}
+		return err
+	}
+	if inspect.State.Running {
```
tomi, line 190: The state [can also be](https://www.baeldung.com/ops/docker-container-states) `created`, `restarting`, `exited`, `paused` or `dead`. I guess `paused` should never happen since we are not pausing them, but the others might happen, especially restarting, so would be nice to handle them gracefully. This can also be a followup no need to fix in this PR

internal/runner/manager/manager.go
```diff
@@ -157,6 +158,74 @@ func (m *Manager) GetContainerInfo(ctx context.Context, containerID string) (*Co
...
+	}
+	if err := m.docker.startContainer(ctx, containerID); err != nil {
+		return fmt.Errorf("start container: %w", err)
+	}
+	containerIP, err := m.docker.containerIP(ctx, containerID)
+	if err != nil {
+		return err
+	}
+	if err := netrules.ApplyPolicy(containerID, containerIP, m.gatewayIP, daemonPort); err != nil {
+		return fmt.Errorf("apply network rules: %w", err)
```
tomi, line 201: As a followup, if this fails, the container will be running but its network policy will be missing. If I understood this correctly, the next request will see it as running and assume everything is OK

tomi (approved):
Couple comments, few ones that would be nice to fix, but can be done as a followup to get this one in

internal/api/config/config.go
endor: There can be a race where a command still arrives "in time", but the deletion has already started. In order to avoid this, we stop taking commands after `IdleDeleteAfter`, but only actually delete the sandbox after the safety buffer.

internal/api/ttl.go
endor: Basically to avoid the extra work in case the sandbox is old enough to be both stopped and deleted.

PR #42, opened 2026-05-13
Add sysbox setup script for Linux machines

tomi:
## Summary

- Adds `scripts/setup-sysbox.sh` to automate sysbox-ce v0.7.0 installation on supported Linux machines (Ubuntu 18/20/22/24, Debian 10/11, amd64/arm64)
- Validates prerequisites (distro, architecture, kernel > 5.19, Docker), installs dependencies (jq, br_netfilter), downloads the `.deb` with SHA256 verification, and verifies the installation
- Prompts before stopping running Docker containers (required for sysbox install) and exits early if sysbox is already installed

## Test plan

- [ ] Run on a supported Ubuntu/Debian machine and verify end-to-end installation
- [ ] Verify early exit when sysbox is already installed
- [ ] Verify prerequisite check failures on unsupported distro/arch/kernel

---
## Summary by cubic
Adds `scripts/setup-sysbox.sh` to automate installing `sysbox-ce` v0.7.0 on supported Ubuntu/Debian hosts (amd64/arm64). It checks distro/arch, kernel > 5.19, Docker, and sudo; installs `jq` and persists `br_netfilter`; downloads the `.deb` with SHA256; prompts before stopping/removing containers; exits early if already installed; guards `xargs` on empty input; and verifies the `sysbox` service and Docker `sysbox-runc` runtime.

<sup>Written for commit 9715b5a1ac694a3971f996421eea60693737fee1. Summary will update on new commits.</sup>


mfsiega (approved):
Seems reasonable! The only thing I would think about would be a flag like `--dry-run` to only check the pre-reqs but not actually install anything. But I'm not sure how if that's really worth it.

tomi:
> Seems reasonable! The only thing I would think about would be a flag like `--dry-run` to only check the pre-reqs but not actually install anything. But I'm not sure how if that's really worth it.

That's a good idea 👍 I'll do that as afollowup, as I have another PR depending on this one already

PR #53, opened 2026-05-18
fix(runner): tear down network rules after the container is gone

mfsiega:
## Summary
- Swap teardown ordering so the sandbox container is removed *before* its iptables rules come down, closing the window where a running container has no egress drops or daemon-port ingress ACL.
- Introduce `stopAndCleanContainer` helper and use it in `DeleteContainer`, the `CreateContainer` error-cleanup closure, and `reconcileContainers` — three sites that previously duplicated the same teardown ordering.
- If `docker rm -f` fails the helper returns early and leaves the rules in place, so a still-running container stays confined.
- Drops the vestigial `inspectContainer` branch in `reconcileContainers` that ran `Teardown` either way with the same log message.

Refs [CAT-3049](https://linear.app/n8n/issue/CAT-3049/handle-timing-issue-in-container-teardown-network-first-vs-container).

## Test plan
- [x] `make fmt-check`
- [x] `make vet`
- [x] `go test ./internal/runner/manager/...`
- [ ] e2e: create + delete a sandbox, confirm container is gone before iptables chains disappear (manual)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Fixes a teardown race by removing the sandbox container before dropping its iptables rules, so a running container is never left unconfined. Adds a shared helper to enforce this order across create, delete, and reconcile paths.

- **Bug Fixes**
  - Remove container first, then call `netrules.Teardown`; if `docker rm -f` fails, rules stay in place to keep confinement.
  - Closes the timing window in CAT-3049 where rules could be gone while the container was still running.

- **Refactors**
  - Add `stopAndCleanContainer` and use it in `DeleteContainer`, the create error-cleanup path, and `reconcileContainers`; it logs `removeContainer` errors internally, returns the error to callers, and includes a TODO for metrics on teardown failures.
  - Drop the unused reconcile branch that inspected the container and called `Teardown` either way.

<sup>Written for commit eb00ef0d3e467a247e4cc2176edc1f1f14fc7699. Summary will update on new commits. <a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/53?utm_source=github">Review in cubic</a></sup>


internal/runner/manager/manager.go
```diff
@@ -219,21 +216,32 @@ func (m *Manager) DaemonURL(ctx context.Context, sandboxID string) (string, erro
...
+// stopAndCleanContainer removes the container, then tears down its network
+// rules. Order matters: rules must outlive the container so it cannot run
+// unconfined during teardown. The Teardown error is logged and swallowed;
+// the removeContainer error is returned so callers decide whether to bail.
+func (m *Manager) stopAndCleanContainer(ctx context.Context, containerID string) error {
+	if containerID == "" {
+		return nil
+	}
+	if err := m.removeContainer(ctx, containerID); err != nil {
+		return err
```
endor, line 202: In all three places we seem to log this error anyway. Wouldn't it be better to log it here and then places calling this function would only need to decide whether to ignore or bubble up the error?

internal/runner/manager/manager.go
```diff
@@ -219,21 +216,32 @@ func (m *Manager) DaemonURL(ctx context.Context, sandboxID string) (string, erro
...
+	}
+	if err := netrules.Teardown(containerID); err != nil {
+		slog.Warn("teardown network rules", "container_id", containerID, "err", err)
```
endor, line 206: This might in the future be a good place for using metrics to become aware of things breaking. This would decouple having the functionality from implementing alerts based on logs.

endor (dismissed):

internal/runner/manager/manager.go
mfsiega: Good call, done.

internal/runner/manager/manager.go
mfsiega: Agreed.  Left a todo for now since it doesn't seem like there's any metrics to hook into? Let me know if you had something else in mind.

endor (approved):

PR #59, opened 2026-05-19
fix: leftovers from lifecycle management

endor:
## Summary

Leftovers from lifecycle management [PR](https://github.com/n8n-io/n8n-sandbox-service/pull/32):
* Handle other docker states besides `Running` in wake up
* Handle failure to apply netrules in wake up
* Add tests for these cases

---
## Summary by cubic
Fixes sandbox wake-up edge cases and enforces correct cleanup order. We now consider full Docker states, and on wake failure we stop the container first, then tear down `netrules` (keeping rules if stop fails).

- **Bug Fixes**
  - Start containers in `created`/`exited`; block `paused`/`restarting`/`dead`/`removing`.
  - On wake failures (IP lookup, policy apply, daemon wait), stop container, then tear down `netrules`; keep rules if stop fails.
  - Readiness checks use full Docker state, not just `Running` or `Status=running`.
  - Reconcile removes leftovers and tears down rules.

- **Refactors**
  - Introduced `dockerBackend` and `newManager` to inject Docker, `netrules` (`applyPolicy`/`teardownRules`), and `waitForDaemon`.
  - Added unit tests for state predicates and wake-up cleanup order, plus an e2e test for network-detach failures and rule removal.

<sup>Written for commit a45d9047f75d5dc3a15e470d6854384037dd593d. Summary will update on new commits. <a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/59?utm_source=github">Review in cubic</a></sup>


e2e/tests/sandbox-recovery.spec.ts
```diff
@@ -151,4 +206,25 @@ test.describe('Sandbox recovery on runner', () => {
       await deleteSandbox(id);
     }
   });
+
+  test('failed wake after network detach stops container and removes rules', async () => {
```
tomi: I'm not sure this needs to be an e2e test. We are testing the internal implementation details of the runner. I feel an integration or unit test would cover this better (we don't actually have integration tests at the moment)

internal/runner/manager/docker_client.go
```diff
@@ -36,6 +34,14 @@ type containerInspect struct {
 	} `json:"NetworkSettings"`
 }
 
+type containerState struct {
+	Status     string `json:"Status"`
+	Running    bool   `json:"Running"`
+	Paused     bool   `json:"Paused"`
+	Restarting bool   `json:"Restarting"`
+	Dead       bool   `json:"Dead"`
```
tomi, line 42: This is a bit weird data model. Is there a reason why we can't use `state: "running" | "paused" | ..`?

internal/runner/manager/manager.go
```diff
@@ -182,26 +200,43 @@ func (m *Manager) ensureSandboxRunningOnce(ctx context.Context, sandboxID string
...
 	return nil
 }
 
+func (m *Manager) cleanupWakeFailure(containerID string) {
+	cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	if err := m.docker.stopContainer(cleanupCtx, containerID); err != nil {
+		slog.Warn("stop container after wake failure", "container_id", containerID, "err", err)
+	}
+	if err := m.teardownRules(containerID); err != nil {
```
tomi, line 243: I think there is still a non-zero chance that if the container stopping fails (e.g. timeout), we will tear down the network rules and leave the container running without network rules. I think we should only remove them if we have successfully stopped the contaienr

tomi (commented):
Couple comments. I think it's better now, but the whole process still feels a bit brittle and difficult to grasp. I created https://linear.app/n8n/issue/CAT-3164/refactor-sandbox-container-lifecycle-management that we can take a look later on

internal/runner/manager/manager.go
endor: Fixed.

internal/runner/manager/docker_client.go
endor: As discussed in Slack, this is just an internal representation of what Docker gives us.

e2e/tests/sandbox-recovery.spec.ts
endor: Fixed.

tomi (approved):
🚀

PR #72, opened 2026-05-25
fix: simplify mTLS cert setup

tomi:
## Summary

- Replaces `bootstrap-local-mtls.sh` with a portable `bootstrap-mtls.sh` that accepts CLI flags, organizes certs into per-service `api/` and `runner/` subdirectories, and is bundled into the API Docker image for use as an init container.
- Adds a `tls-init` init container to both compose examples (Linux + macOS) so mTLS certs are generated automatically on first run — no manual bootstrap step needed.
- Removes the `--world-readable` flag from compose examples, which was setting private key permissions to 644 (readable by all local users on the host). Instead, uses a targeted `chown` to make API keys readable only by the `sandbox-api` container user.
- Updates e2e scripts, quickstart docs, and `compose.tls.yaml` to use the new script and per-service directory layout.

https://linear.app/n8n/issue/CAT-3205

## Review / Merge checklist

- [ ] I have seen this code, I have run this code, and I take responsibility for this code.
- [ ] PR title and summary are descriptive.
- [ ] Tests included.


endor (commented):
It seems to me like there's quite the overlap between OS X and Linux in terms of documentation/setup. Would it make sense to merge them and have a short paragraph explaining the differences instead?

endor (approved):

tomi:
> It seems to me like there's quite the overlap between OS X and Linux in terms of documentation/setup. Would it make sense to merge them and have a short paragraph explaining the differences instead?

There is indeed but that has been intentional. IMHO it's easier to follow a guide that doesn't have conditional logic in it. I can see if it makes sense to unify them

PR #91, opened 2026-05-29
Add retry-aware exec stream proxy

tomi:
## Summary

Load testing the service showed that under load the execution requests start dropping. The client receives a `started` event, then `unexpected EOF`. This happens in the communication between the runner and the daemon. This can be due to various reasons.

This PR adds retrying onto the runner's proxy. We already have retrying in the SDK, but it is faster to do it on the runner side, instead of cascading the entire request chain.

- Adds a retry-aware runner exec proxy that generates and propagates exec IDs, exposes `X-Exec-Id`, and resumes interrupted daemon NDJSON streams from the last sequence.
- Restores daemon-side JSON body limits and hardens stream parsing and write error handling with targeted tests.
- Documents resume behavior in the API and architecture docs.

https://linear.app/n8n/issue/CAT-3270/retry-aware-exec-proxy-resume-truncated-ndjson-streams-using-daemons

## Review / Merge checklist

- [x] Tests included.

Tests: `make fmt-check`, `make vet`, `go test ./...`.

---
## Summary by cubic
Adds a retry-aware exec proxy for `POST /sandboxes/{id}/executions` that resumes truncated NDJSON streams from the daemon with exponential backoff, reducing failures from transient drops. Exposes `X-Exec-Id` so clients can resume too; implements Linear CAT-3270.

- **New Features**
  - Auto-resume via `GET /executions/{exec_id}?follow=true&after=<seq>` (up to 3 retries, exponential backoff starting at 50 ms).
  - Always sets/propagates `exec_id` and returns it in `X-Exec-Id`.
  - Preserves NDJSON shape and per-line flushing; tracks `seq` via fast prefix parsing.

- **Bug Fixes**
  - Enforces JSON body size limits on runner and daemon endpoints; returns clear 400s for oversized payloads.
  - Hardens stream parsing and write error handling; ensures daemon events start with `seq` and `type` for reliable resume.
  - Stabilizes the public internet e2e check by using `curl` against a stable HTTPS endpoint.

<sup>Written for commit b17f8bb32a8c7877046fa57d37f4091d0e29cdf8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/91?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


e2e/tests/network-isolation.spec.ts
```diff
@@ -2,13 +2,28 @@ import { test, expect } from '@playwright/test';
 import './matchers';
 import { createSandbox, deleteSandbox, exec, execWithTransientRetry } from './helpers';
 
-// Helper: use python3 for HTTP requests since curl is not in the sandbox rootfs
```
endor, line 5: I think some sort of comment should stay for why we need python for this. Ideally we wouldn't. Could we achieve the same thing with some other tool available in the sandbox? nc? node?

endor (dismissed):

mfsiega (dismissed):
LGTM!

e2e/tests/network-isolation.spec.ts
tomi: Switched over to curl

endor (approved):
I think this is good to go. Might be worth it to remove the whole pyConnect stuff in another PR.

tomi:
> I think this is good to go. Might be worth it to remove the whole pyConnect stuff in another PR.

Oh right there are those as well. Yeah let's clean then up separately

tomi:
https://linear.app/n8n/issue/CAT-3327/remove-pyconnect-code

PR #96, opened 2026-06-08
feat: firecracker runner e2e tests

endor:
## Summary

Add e2e tests for the firecracker runner using the same pattern we have in sysbox: dynamically create a VM before and tear it down after running the tests.

---
## Summary by cubic
Adds end-to-end tests for the Firecracker runner. Provisions an Azure VM, prepares Firecracker assets (KVM preflight, install binaries, build kernel/rootfs and a local snapshot from pinned Firecracker CI inputs plus a locally built daemon), starts API and runner on-host, and runs tagged Playwright specs; includes a CI workflow gated by the `e2e-firecracker` label with diagnostics and teardown.

- New Features
  - Adds `.github/workflows/e2e-firecracker.yml` to run Firecracker e2e on manual dispatch or PRs labeled `e2e-firecracker`; provisions an Azure VM via Terraform (OIDC), runs tests over SSH, uploads artifacts on failure, and cleans up.
  - Introduces Firecracker setup: `provision-firecracker-e2e-vm.sh`, `setup-firecracker-e2e-vm.sh`, `create-golden-snapshot.sh` perform KVM checks, install Firecracker/jailer, build an ext4 rootfs from Firecracker CI Ubuntu squashfs, inject the sandbox daemon, and create the snapshot locally; configurable via env (region/VM size, `FIRECRACKER_VERSION`/`FIRECRACKER_CI_VERSION` with optional `FIRECRACKER_TARBALL_SHA256`, rootfs/snapshot sizes). Writes `/srv/firecracker/manifest.json` and adds a failure log collector. Terraform supports a region override via `location`; transfer scripts exclude macOS metadata.
  - Adds a Firecracker runner binary and image: `cmd/runner-firecracker`, `Dockerfile.runner-firecracker`, and `make docker-firecracker-runner-amd64`; introduces a Firecracker runtime (`internal/runner/runtime/firecracker/*`) with per-slot netns + TAP and a host TCP proxy with time-bounded dials, plus unit tests and README.
  - E2E harness: `e2e/run-firecracker.sh` and `e2e/run-firecracker-azure.sh` start API and the Firecracker runner on-host, and select tests by tag. Specs use `tests/tags.ts` to mark lanes (`@firecracker-runner`, `@docker-runner`); `e2e/run.sh` now filters Docker lane tests via tag.

- Bug Fixes
  - Runtime capacity keeps a slot reserved until the per-sandbox proxy fully stops; capacity is released only after host cleanup completes. Adds a unit test.
  - Daemon hardening: drops to a sandbox user at startup when run as root, executes child processes with sandbox credentials, and sets `HOME`. Logs scanner errors in daemon I/O.
  - Infra: packaging excludes macOS metadata; `.github/workflows/e2e-sysbox-sweep.yml` updates `azure/login` to `v3`.

<sup>Written for commit c8ed1005f02ad49ca998775114721386876779bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/96?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


e2e/infra/scripts/setup-firecracker-e2e-vm.sh
```diff
@@ -0,0 +1,305 @@
...
+	if command -v file >/dev/null 2>&1; then
+		file -b "$1"
+	fi
+}
+
+json_escape() {
+	node -e 'process.stdout.write(JSON.stringify(process.argv[1] ?? ""))' "$1"
+}
+
+s3_latest_key() {
```
tomi, line 179: This is not at least for me clear what S3 key we are fetching here? Could we document it?

e2e/tests/sandbox-api.spec.ts
```diff
@@ -13,8 +13,9 @@ import {
 import { SandboxServiceError } from '@n8n/sandbox-client';
 
 const API_KEY = process.env.SANDBOX_API_KEY || 'test';
+const BOTH_RUNNERS = { tag: ['@docker-runner', '@firecracker-runner'] };
```
tomi: This is used only in this suite file. Perhaps we could extract this into some shared place and use everywhere? So something like

```ts
export const filterToRunner: {
  docker: { tag: ['@docker-runner'] },
  firecracker: { tag: ['@firecracker-runner'] },
  both: { tag: ['@docker-runner', '@firecracker-runner'] },
}
```

and then use it everywhere

```ts
test.describe('Auth', filterToRunner.both, () => {
```

internal/daemon/files.go
```diff
@@ -123,11 +123,17 @@ func HandleFileWrite(path string, data []byte, maxBytes int64, overwrite bool) e
 	if err := os.MkdirAll(dir, 0o755); err != nil {
 		return fmt.Errorf("file_write mkdir %s: %w", dir, err)
 	}
+	if err := chownSandboxPath(dir); err != nil {
```
tomi: We didn't need these before, why do we need them now?

e2e/infra/variables.tf
```diff
@@ -8,6 +8,12 @@ variable "vm_name" {
   type        = string
 }
 
+variable "location" {
+  description = "Azure region for VM resources."
+  type        = string
+  default     = "germanywestcentral"
```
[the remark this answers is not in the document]
endor: This is fine. We're using SKU's that are compatible with gwc.

internal/daemon/files.go
endor: These were added when the daemon as still running as root in the Firecracker VM. I originally wanted to make a separate PR/change to run is as non-root, but now this has already ended up in this PR, so these shouldn't be needed anymore. Will remove.

e2e/tests/sandbox-api.spec.ts
endor: Fixed.

e2e/infra/scripts/setup-firecracker-e2e-vm.sh
endor: Fixed. This is about fetching the latest Firecracker assets from AWS: the kernel image, rootfs.

e2e/tests/file-exec-consistency.spec.ts
```diff
@@ -16,7 +16,7 @@ import {
  */
 test.describe.configure({ mode: 'serial' });
 
-test.describe('File API and Exec API path consistency', { tag: ['@docker-runner', '@firecracker-runner'] }, () => {
+test.describe('File API and Exec API path consistency', { tag: '@both' }, () => {
```
tomi: Sorry if I was unclear, but this is not really what I meant. IMO it makes sense to have separate tags, just unify how they are specified so we don't have magic strings but rather references. Or does that not work with playwright for some reason?

endor: Fixed.

tomi (approved):
🚀

PR #104, opened 2026-07-17
feat: add basic tenant management

endor:
## Summary

* `SANDBOX_API_KEYS` are considered admin keys. 
* Admin keys can be used to create tenants and mint tenant api keys.
* Tenant API keys can do normal sandbox operations but will only be able to access the sandboxes they created.
* e2e and smoke tests use tenant API keys.
* We keep backwards compatibility by allowing `SANDBOX_API_KEYS` to be used as before.

---
## Summary by cubic
Adds basic tenant management to the Sandbox API. Admin keys can create tenants and mint tenant-scoped keys; tenant keys are limited to their own sandboxes with per-tenant quotas and 404s for cross-tenant access.

- **New Features**
  - `/admin/tenants`: create/list/get/delete tenants and mint/list/revoke keys (plaintext key returned once); deleting a tenant with sandboxes returns `409`.
  - DB-backed tenant auth: keys stored hashed with prefix lookup; `/admin/*` requires admin (`403` for tenant keys). List/get/delete/proxy are scoped to the caller; cross-tenant sandboxes return `404`.
  - Sandbox creation records `tenant_id` and enforces per-tenant `max_sandboxes` (`0` = unlimited; soft check-then-act before create).
  - Config and schema: `SANDBOX_API_DEFAULT_MAX_SANDBOXES` and Helm `api.config.defaultMaxSandboxes` (default 50); migrations add `tenants`, `api_keys`, and `sandboxes.tenant_id` (SQLite/Postgres).
  - Tests/tooling: e2e and smoke mint a tenant key with the admin key and run with it; smoke cleans up the tenant.
  - SDK: `@n8n/sandbox-client` `deleteSandbox` treats HTTP `404` as success to keep deletes idempotent.

- **Migration**
  - No action needed; `SANDBOX_API_KEYS` remain admin keys with full access. Note: DELETE of a missing sandbox now returns `404`.
  - To adopt tenants, create one via `/admin/tenants` and store the returned plaintext key.
  - Optionally set per-tenant quotas via `SANDBOX_API_DEFAULT_MAX_SANDBOXES` or Helm `api.config.defaultMaxSandboxes`.
  - CI/smoke should provide an admin key; the scripts will mint a tenant key automatically.

<sup>Written for commit 1790b6966467694d9dd11d071c323c961a0cca85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


internal/api/handlers.go
```diff
@@ -165,6 +178,36 @@ func handleCreateSandbox(s store.SandboxStore, reg registry.RunnerRegistry, cfg
...
+			if err != nil {
+				writeError(w, http.StatusInternalServerError, err.Error())
+				return
+			}
+			if tenant == nil {
+				writeError(w, http.StatusUnauthorized, "invalid API key")
+				return
+			}
+			if tenant.MaxSandboxes > 0 {
+				n, err := s.CountByTenant(tenantID)
```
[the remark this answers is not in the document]
endor: This is fine. We'll treat `MaxSandboxes` as a soft limit for now, so we won't have to deal with the complexity and performance penalty of adding a lock.

tomi (commented):
Looks mostly good, nothing major to comment 👍 Few minor comments and questions

docs/API.md
```diff
@@ -602,3 +619,107 @@ Get file or directory metadata.
...
+  "create_key": true
+}
+```
+
+| Field          | Type   | Required | Default |
+|----------------|--------|----------|---------|
+| `name`         | string | no       | `""` |
+| `external_ref` | string | no       | `""` |
+| `max_sandboxes`| int    | no       | `SANDBOX_API_DEFAULT_MAX_SANDBOXES` (`50`) |
+| `create_key`   | bool   | no       | `true` |
```
tomi, line 682: Is there a use case for creating tenants without an API key?

internal/api/handlers_tenants.go
```diff
@@ -0,0 +1,322 @@
...
+		createKey := true
+		if req.CreateKey != nil {
+			createKey = *req.CreateKey
+		}
+
+		now := time.Now().Unix()
+		t := &store.Tenant{
+			ID:           uuid.New().String(),
+			Name:         req.Name,
+			ExternalRef:  req.ExternalRef,
```
tomi, line 123: nit: We should validate that these don't bypass the limit

internal/api/store/store.go
```diff
@@ -24,9 +31,29 @@ type SandboxRecord struct {
...
+}
+
+// APIKey is a hashed tenant credential. Plaintext is never stored.
+type APIKey struct {
+	ID        string
+	TenantID  string
+	KeyHash   string
+	Prefix    string
+	CreatedAt int64
+	RevokedAt int64 // 0 = active
```
tomi, line 73: Do we remove these ever?

internal/api/store/migrations_sqlite.go
```diff
@@ -24,3 +24,26 @@ const sqliteAddRunnerIDCol = `ALTER TABLE sandboxes ADD COLUMN runner_id TEXT DE
 const sqliteAddRunnerHTTPBaseURLCol = `ALTER TABLE sandboxes ADD COLUMN runner_http_base_url TEXT DEFAULT ''`
 
 const sqliteAddRunnerControlGRPCAddrCol = `ALTER TABLE sandboxes ADD COLUMN runner_control_grpc_addr TEXT DEFAULT ''`
+
+const sqliteAddTenantIDCol = `ALTER TABLE sandboxes ADD COLUMN tenant_id TEXT NOT NULL DEFAULT ''`
```
tomi: Does `''` mean it is owned by admin? Should we have some special marker for these?

internal/api/store/postgres.go
```diff
@@ -53,12 +65,14 @@ func (s *PostgresStore) Backend() Backend { return BackendPostgres }
 
 func (s *PostgresStore) Close() error { return s.db.Close() }
 
+const pgSandboxCols = `id, status, created_at, last_active_at, rootfs_path, socket_path, container_ip, daemon_port, runner_id, runner_http_base_url, runner_control_grpc_addr, tenant_id`
```
tomi, line 87: This is nicer 👍

e2e/tests/helpers.ts
```diff
@@ -4,22 +4,65 @@ import * as https from 'node:https';
 import { SandboxClient, type ExecResult } from '@n8n/sandbox-client';
 import { execFileSync } from 'node:child_process';
 
-const API_KEY = process.env.SANDBOX_API_KEY || 'test';
-const BASE_URL = process.env.BASE_URL || 'http://localhost:8080';
+/** Admin key from env (SANDBOX_API_KEYS). Used to mint tenant keys. */
+export const ADMIN_API_KEY = process.env.SANDBOX_API_KEY || 'test';
+const BASE_URL = process.env.BASE_URL || process.env.BASE_URL_A || 'http://localhost:8080';
```
tomi, line 9: What's `BASE_URL_A`?

e2e/tests/helpers.ts
```diff
@@ -4,22 +4,65 @@ import * as https from 'node:https';
...
+  }
+  return tenantMintPromise;
+}
+
+export async function ensureTenantAuth(mintBaseUrl?: string): Promise<void> {
+  await getApiKey(mintBaseUrl);
+}
+
+/** Tenant-scoped client (initialized after ensureTenantAuth / first helper call). */
+export let client = new SandboxClient({ baseUrl: BASE_URL, apiKey: ADMIN_API_KEY });
```
tomi: The doc and the default value doesn't really match, as it is initialized via the ADMIN_API_KEY

e2e/tests/helpers.ts
```diff
@@ -4,22 +4,65 @@ import * as https from 'node:https';
...
+
+export function sandboxClient(baseUrl: string = BASE_URL, apiKey?: string): SandboxClient {
+  return new SandboxClient({ baseUrl, apiKey: apiKey ?? tenantApiKey ?? ADMIN_API_KEY });
 }
```
tomi: This is a bit convoluted. Could we have separate functions for adminClient and tenantClient and client with given API key? Now it could be any depending on state, and it would be cleaner if these were separate methods

e2e/tests/sandbox-api.spec.ts
```diff
@@ -759,6 +762,68 @@ test.describe('Deleted Sandbox 404 Tests', () => {
   });
 });
 
+test.describe('Tenant isolation', () => {
+  test('tenant cannot access another tenant sandbox', async ({ request }) => {
```
tomi, line 818: Would be nice to have a test to also check that another tenant cannot access other tenant's sandbox via the file or exec endpoinst

internal/api/store/postgres.go
```diff
@@ -168,30 +177,151 @@ func (s *PostgresStore) Count() (int64, error) {
...
+	return out, rows.Err()
+}
+
+func (s *PostgresStore) DeleteTenant(id string) error {
+	n, err := s.CountByTenant(id)
+	if err != nil {
+		return err
 	}
-	return records, nil
+	if n > 0 {
```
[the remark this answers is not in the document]
tomi: This is a valid race condition. We could acquire a DB lock here to serialize this with sandbox creation

docs/API.md
endor: Only if the process should be to first create the tenant and then mint keys later (maybe as part of a different workflow). We could remove this property and always create it or not create it and always require minting.

internal/api/handlers_tenants.go
endor: Sorry, which limit are we talking about?

internal/api/store/store.go
endor: Not at the moment. Slim down or leave for later?

e2e/tests/helpers.ts
endor: It represents the "first" pod in a multi-pod setup.

internal/api/store/migrations_sqlite.go
endor: Fixed.

e2e/tests/helpers.ts
endor: Fixed.

e2e/tests/helpers.ts
endor: Fixed.

e2e/tests/sandbox-api.spec.ts
endor: Fixed.

internal/api/handlers_tenants.go
tomi: Oh looks like these don't have any limits on DB side. That's surprising. Would be nice to have at least some max limit for `name` and `external_ref` instead of being completely unbounded

internal/api/store/store.go
tomi: I guess it would make sense to remove them directly when revoked?

internal/api/store/postgres.go
```diff
@@ -94,6 +102,38 @@ func (s *PostgresStore) Create(record *SandboxRecord) error {
...
+	}
+
+	tx, err := s.db.Begin()
+	if err != nil {
+		return fmt.Errorf("store: begin create sandbox: %w", err)
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	// Lock the tenant row so DeleteTenant cannot commit until this insert finishes.
+	res, err := tx.Exec(`UPDATE tenants SET name = name WHERE id = $1`, record.TenantID)
```
tomi: No need do an artificial update, we can use [`SELECT ... FOR UPDATE`](https://www.postgresql.org/docs/18/explicit-locking.html#LOCKING-ROWS), which locks the row:

```sql
SELECT id FROM tenants WHERE id = $1 FOR UPDATE
```

internal/api/store/postgres.go
```diff
@@ -259,17 +299,40 @@ func (s *PostgresStore) ListTenants() ([]*Tenant, error) {
...
+	tx, err := s.db.Begin()
 	if err != nil {
-		return err
+		return fmt.Errorf("store: begin delete tenant: %w", err)
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	// Lock the tenant row so a concurrent Create cannot insert a sandbox after
+	// the emptiness check and before this delete commits.
+	res, err := tx.Exec(`UPDATE tenants SET name = name WHERE id = $1`, id)
```
tomi: Same here, we can use SELECT FOR UPDATE

internal/api/store/sqlite.go
```diff
@@ -119,6 +128,38 @@ func (s *SQLiteStore) Create(record *SandboxRecord) error {
 	return nil
 }
 
+func (s *SQLiteStore) Create(record *SandboxRecord) error {
+	if IsAdminTenantID(record.TenantID) {
+		return s.insertSandbox(s.db, record)
+	}
+
+	tx, err := s.db.Begin()
```
tomi: If we open the transaction with [`BEGIN IMMEDIATE`](https://sqlite.org/lang_transaction.html), it will lock the DB file so other connections can't write it. With that we don't need the artificial `UPDATE` based locking

internal/api/store/sqlite.go
```diff
@@ -186,28 +243,187 @@ func (s *SQLiteStore) Count() (int64, error) {
...
+	const tenantQ = `INSERT INTO tenants (id, name, external_ref, max_sandboxes, created_at) VALUES (?, ?, ?, ?, ?)`
+	if _, err := tx.Exec(tenantQ, t.ID, t.Name, t.ExternalRef, t.MaxSandboxes, t.CreatedAt); err != nil {
+		return fmt.Errorf("store: create tenant %s: %w", t.ID, err)
+	}
+	var revoked any
+	if k.RevokedAt > 0 {
+		revoked = k.RevokedAt
+	}
+	const keyQ = `INSERT INTO api_keys (id, tenant_id, key_hash, prefix, created_at, revoked_at) VALUES (?, ?, ?, ?, ?, ?)`
+	if _, err := tx.Exec(keyQ, k.ID, k.TenantID, k.KeyHash, k.Prefix, k.CreatedAt, revoked); err != nil {
```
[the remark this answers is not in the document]
tomi: Valid point ☝️ I think we should assert that `t.ID == k.TenantID`

internal/api/store/sqlite.go
```diff
@@ -284,17 +325,40 @@ func (s *SQLiteStore) ListTenants() ([]*Tenant, error) {
 }
 
 func (s *SQLiteStore) DeleteTenant(id string) error {
-	n, err := s.CountByTenant(id)
+	tx, err := s.db.Begin()
```
tomi: Same here, let's use `BEGIN IMMEDIATE`

tomi (commented):
Few more comments

internal/api/handlers_tenants.go
tomi: Then again since this is admin only endpoint having unbounded input is not the worst that can happen, so I think it's fine

internal/api/store/store.go
endor: Once we have audit logs, I agree. Until then it might be better to keep them around as revoked. We already have a Linear issues for audit logs (even though it's currently deprioritized to M4, but we could move it up).

tomi (approved):
LGTM 🚀

PR #124, opened 2026-08-17
fix: serialize firecracker sandbox lifecycle operations per sandbox

endor:
## Summary

The Firecracker runner only guarded stop/wake against each other, so a delete could run
concurrently with an in-flight create or wake and tear down a sandbox while another
goroutine was still setting up its slot, netns and socket paths. Replaced the `stopping`
flag with a `transition` field that all four lifecycle operations claim, so create, stop,
wake and delete are serialized per sandbox while still running in parallel across
different sandboxes.

Also fixes an unsynchronized read of `len(r.sandboxes)` in the snapshot disk check, adds
`-race` to `make test`, and adds regression tests for delete-during-create and
delete-during-wake plus an e2e spec that races a delete against a wake.

---
## Summary by cubic
Serializes Firecracker sandbox lifecycle per sandbox so delete cannot tear down around in‑flight create/wake and orphan microVMs. Previously only stop/wake were mutually exclusive; now create/stop/wake/delete claim a per‑sandbox transition, delete waits for create/wake, and waiting is bounded.

- Adds a transition enum and `beginTransition`/`endTransition` plus lifecycle budgets: create (3m) and stop/wake/delete (2m). Introduces a separate wait budget (create + transition) so waiting does not eat the winner’s budget; operations run detached from caller cancellation.
- CreateSandbox claims `transitionCreating` in `reserveSandbox` and clears it after publish. `DaemonURL` now rejects while any transition is in flight, when stopped, or before publish during create; callers retry on `ErrSandboxNotRunning`.
- Wake failures clean up host state on a fresh cleanup budget even if the wake’s own budget expired, preventing leaked mounts/netns before the slot is released.
- Fixes a race in snapshot disk checks by reading sandbox count under lock; `make test` now runs `go test -race`.
- Eviction skips sandboxes with any in‑flight transition (including waking) and marks `transitionDeleting` when evicting.
- Shutdown does not wait for stop/wake; it marks `transitionDeleting` and skips sandboxes already claimed for delete.
- Tests cover: delete during create and wake, delete waiting out a long create, bounded waits for held transitions, delete after caller cancel, wake failure cleanup, shutdown vs in‑flight delete, and an e2e spec racing delete vs wake.

<sup>Written for commit 6bfe9760bab5895030578a93331e443d36fd7aa3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/124?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tomi (commented):
Couple questions. AI also flagged these:

```
[P2] Deletion claim is not atomic — runtime.go:416

beginTransition releases r.mu before state.deleting is set. If Shutdown runs in that gap—possible while the unjoined admission Prepare goroutine is cleaning up—it sees deleting == false, selects the same state, and both paths execute deleteSandbox concurrently. This can double-stop resources and race host cleanup. Set deleting in the same critical section as the transition claim, or make Shutdown skip an existing deletion transition.

[P2] Regression test cannot observe the targeted leak — sandbox-delete-during-wake.spec.ts:48

The old broken path removes the sandbox state and releases its slot before the wake publishes its process/proxy, so both gauges return to baseline even if host resources are subsequently orphaned; the DELETE and 404 assertions also succeed. This test can therefore pass against the regression it is intended to catch. Verify actual Firecracker host resources or successfully reuse the released slot with a replacement sandbox.
```

internal/runner/runtime/firecracker.ee/runtime.go
```diff
@@ -315,8 +331,9 @@ func (r *Runtime) DaemonURL(_ context.Context, sandboxID string) (string, error)
 	if !ok || state.deleting {
 		return "", runnerruntime.ErrSandboxNotFound
 	}
-	if state.stopping || state.stopped || !state.running {
-		// Reject proxies while stopping, stopped, or not yet running (create in progress).
+	if state.transition != transitionNone || state.stopped || !state.running {
```
tomi, line 408: Ideally if the sandbox is still being created / waken up, we'd like to make any requests wait until it is up. Do we handle this error somewhere and wait/retry? E.g. in the SDK or API

internal/runner/runtime/firecracker.ee/runtime.go
```diff
@@ -384,16 +407,19 @@ func (r *Runtime) reserveSandbox(sandboxID string) (*sandboxState, error) {
...
 
-// takeSandbox marks the sandbox as deleting while keeping its slot reserved
-// until host cleanup finishes.
-func (r *Runtime) takeSandbox(sandboxID string) (*sandboxState, error) {
+// takeSandbox waits for any in-flight stop or wake to finish, then marks the
+// sandbox as deleting while keeping its slot reserved until host cleanup
+// finishes. Waiting is what guarantees teardown sees the microVM a concurrent
+// wake was bringing up instead of leaving it orphaned on a slot the runner has
+// already handed back.
+func (r *Runtime) takeSandbox(ctx context.Context, sandboxID string) (*sandboxState, error) {
```
tomi: This seems to be for delete case specifically. Maybe it could be name takeSandboxForDeletion?

internal/runner/runtime/firecracker.ee/stop_wake.go
```diff
@@ -211,60 +205,57 @@ func (r *Runtime) teardownRunningVM(ctx context.Context, state *sandboxState) er
...
-func (r *Runtime) waitWhileStopping(ctx context.Context, sandboxID string) error {
-	ticker := time.NewTicker(50 * time.Millisecond)
+// beginTransition waits until no other stop, wake, or delete is in flight for the
+// sandbox, then claims it for the caller. Callers must release the claim with
+// endTransition unless the transition is terminal (delete). The poll loop keeps
+// waiters off r.mu while a transition runs host commands.
+func (r *Runtime) beginTransition(ctx context.Context, sandboxID string, want sandboxTransition) (*sandboxState, error) {
+	ticker := time.NewTicker(transitionPollInterval)
 	defer ticker.Stop()
 	for {
```
tomi, line 231: Do all operations have a bound time limit? Or should this have a timeout?

e2e/tests/sandbox-delete-during-wake.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const id = await createSandbox();
+    await execWithTransientRetry(id, 'echo before-stop');
+    stopSandboxViaRunner(id);
+
+    // The exec triggers a wake and the delete lands while it is in flight. Either
+    // may win, so the exec outcome is deliberately not asserted; what matters is
+    // that the delete settles and the runner is left clean. A leaked transition
+    // claim would instead hang the delete until the test timeout.
+    const results = await Promise.allSettled([exec(id, 'echo racing-wake'), deleteSandbox(id)]);
```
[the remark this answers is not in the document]
endor: Irrelevant for e2e tests, as that's cleaned up after the test run anyway by dropping the runner.

Makefile
```diff
@@ -50,9 +50,9 @@ api:
 playground: sdk-install sdk-build
 	cd playground && npm install && npm start
 
-## test: Run all tests.
+## test: Run all tests with the race detector.
 test:
-	go test ./...
+	go test -race ./...
```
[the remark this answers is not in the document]
endor: I think this is fine for now.

internal/runner/runtime/firecracker.ee/runtime.go
endor: What we currently have:

1. concurrent wakes: `wakeGroup` makes them wait
2. concurrent creates with the same uuid: they're serialized on the per-sandbox lock and return 201, 200, 200, ..
3. exec before sandbox is created: 404
4. exec while create is in flight: 404 until record is in api db

internal/runner/runtime/firecracker.ee/stop_wake.go
endor: Fixed.

internal/runner/runtime/firecracker.ee/stop_wake.go
```diff
@@ -85,40 +82,35 @@ func (r *Runtime) EnsureSandboxRunning(ctx context.Context, sandboxID string) er
...
 	})
 	return err
 }
 
 func (r *Runtime) ensureSandboxRunningOnce(ctx context.Context, sandboxID string) error {
-	if err := r.waitWhileStopping(ctx, sandboxID); err != nil {
-		return err
-	}
-	state, err := r.sandboxForLifecycle(sandboxID)
+	state, ctx, cancel, err := r.beginTransition(ctx, sandboxID, transitionWaking)
```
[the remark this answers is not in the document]
endor: Fixed.

internal/runner/runtime/firecracker.ee/runtime.go
```diff
@@ -78,6 +78,62 @@ func (s slotState) occupied() bool {
...
+// API exits), so a wedged host command would otherwise hold the claim forever and
+// fail every later operation on that sandbox.
+//
+// beginTransition applies transitionBudget to stop, wake, and delete. Create has
+// its own, larger budget because it clones the rootfs and snapshot before booting,
+// and it claims the sandbox in reserveSandbox rather than through beginTransition.
+// Vars rather than consts so tests can shrink them.
+var (
+	createBudget     = 3 * time.Minute
+	transitionBudget = 2 * time.Minute
```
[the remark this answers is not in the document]
endor: Fixed.

internal/runner/runtime/firecracker.ee/stop_wake.go
```diff
@@ -192,60 +186,57 @@ func (r *Runtime) teardownRunningVM(ctx context.Context, state *sandboxState) er
...
-		if !ok {
-			return runnerruntime.ErrSandboxNotFound
+		if !ok || state.deleting {
+			r.mu.Unlock()
+			return nil, runnerruntime.ErrSandboxNotFound
 		}
-		if !stopping {
-			return nil
+		if state.transition == transitionNone {
+			state.transition = want
```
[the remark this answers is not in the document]
endor: Fixed.

internal/runner/runtime/firecracker.ee/runtime.go
tomi: I think that's fine, as usually the flow is to first create sandbox, then exec. And if an exec waits for a waking up sandbox it's all good

internal/runner/runtime/firecracker.ee/runtime.go
```diff
@@ -93,6 +97,62 @@ const (
...
+)
+
+// transitionWaitBudget bounds how long an operation waits for another one to
+// release the sandbox. It is separate from, and longer than, the budget the winner
+// then runs under, for two reasons: waiting must not eat the time the operation
+// needs for its own host work, and a waiter must not give up on a claim that is
+// still guaranteed to be released. The longest possible hold is a create that
+// exhausts createBudget and then runs its cleanup, hence the sum.
+func transitionWaitBudget() time.Duration {
+	return createBudget + transitionBudget
```
tomi, line 124: 5mins is pretty long. Is that realistic that things can take that long?

tomi (approved):
LGTM. One comment about the max time of waits

internal/runner/runtime/firecracker.ee/runtime.go
endor: No, it's not. It's an edge case. I'd suggest we leave it as is for now and later on can try to make it smaller if necessary.

PR #126, opened 2026-08-19
feat(chart): support clusters that cannot install sysbox

### LINKED ISSUE #122: In-cluster data plane is unreachable on immutable-rootfs Kubernetes distributions, because sysbox is the only supported mode
**TL;DR** — `dataPlane.mode: sysbox` is the only in-cluster option, and sysbox cannot be installed on an immutable-rootfs distribution, so those clusters have no in-cluster data plane at all. We hit this on Talos; the same constraint should apply to Flatcar, Fedora CoreOS and Bottlerocket, though we have only tested Talos. Since `sysboxRunner.image.repository` is already `n8n-sandbox-service-runner-dind`, we added a third mode that runs the same image privileged instead of under sysbox, and it has been running on Talos. We would rather contribute it than carry a fork: what shape would you accept?

The chart's `dataPlane.mode` accepts `sysbox` or `external`, and the validation template fails the render on anything else. The in-cluster path therefore requires a Sysbox-ready node pool, which the README lists as a prerequisite.

That prerequisite cannot be satisfied on a growing class of Kubernetes distributions, and we do not think the reason is a niche one.

## Why the prerequisite cannot be met

Sysbox installs by placing a runtime binary on the host and editing the node's containerd configuration to register `sysbox-runc`. Both steps mutate the node filesystem.

Immutable-rootfs distributions do not permit that. Talos has no shell, no package manager, and a read-only, verity-protected root; the containerd configuration is machine-config-managed rather than file-managed. The same constraint applies in kind to Flatcar, Fedora CoreOS and Bottlerocket. On managed control planes where node images are vendor-owned, installing a third-party runtime is at best unsupported and often prohibited by policy.

So the requirement is not "add a node pool". It is "run a different operating system".

## Why this reads as an enterprise problem rather than a hobbyist one

The distributions that block this are the ones chosen specifically for security posture. Immutability, verified boot, no interactive access, and a minimal attack surface are the reasons an organisation picks Talos or Bottlerocket in the first place, and they are frequently the reasons a compliance regime accepts the cluster at all.

The result is inverted from what the design presumably intends: the more hardened the platform, the less able it is to run the sandbox in-cluster. Teams that have invested in node-level hardening are the ones told they cannot use the isolation feature.

It also affects exactly the deployments most likely to want a self-hosted sandbox rather than a managed one, since the motivation for self-hosting is usually that code execution must stay inside a controlled boundary.

## The failure is legible, but not where you look for it

Worth separating from the design question, because it costs time regardless of what you decide.

Installing with the documented defaults on a cluster without sysbox produces a successful `helm install`. The StatefulSet is created and then no pod ever appears, because the RuntimeClass admission plugin rejects it:

```
pods "<release>-sysbox-runner-0" is forbidden: pod rejected: RuntimeClass "sysbox-runc" not found
```

The message is exact and names the cause, so the problem is placement rather than absence: it lands on the StatefulSet's events, and `kubectl get pods` shows nothing at all. The visible state is a chart that installed cleanly and a data plane that never registers, and the first symptom most operators meet is in the n8n UI. The default `sysboxRunner.scheduling.nodeSelector` of `sysbox-install: "yes"` is a second, independent blocker on the same path.

A line in the chart README would cover it: install succeeds, no runner pod is created, check `kubectl describe statefulset`. A render-time preflight is tempting, but `lookup` returns empty under `helm template` and `--dry-run`, so failing on it would break plain templating and GitOps diffs. `NOTES.txt` is the safer place if you want it in the chart itself.

## `external` mode is not a substitute

It is the documented alternative, but it changes the deployment model rather than the runtime: runner hosts move outside Kubernetes and become separately provisioned, separately patched machines. For an organisation running Kubernetes precisely so that workloads are not pets on VMs, that is a larger concession than the sandbox feature is worth. The Firecracker runner is also external-host only and not charted, so it does not close the gap either.

## What we found while working around it

The two in-cluster runners are closer than the mode names suggest. `sysboxRunner.image.repository` is `n8nio/n8n-sandbox-service-runner-dind` — the sysbox mode already runs Docker-in-Docker. What sysbox supplies is not a different workload but a different source of privilege: user-namespace remapping from the runtime, instead of `privileged: true` from the kernel.

We are not proposing that as a default, and we would not argue it is equivalent. `privileged: true` is a materially weaker boundary than sysbox: it drops the user-namespace remapping that makes sysbox worth using, and a container escape reaches the node. It is a trade-off an operator should have to opt into explicitly, with the cost stated plainly, on a namespace with Pod Security Admission set to `privileged`. It is also not a universal answer, since it does nothing for GKE Autopilot, which blocks privileged containers outright.

The point is narrower: the workload already is DinD, so supporting it without sysbox is a question of where privilege comes from, not a new isolation model.

### How we resolved it

We added `dataPlane.mode: dind` as a third mode, in a fork at [TpyoKnig/n8n-sandbox-service](https://github.com/TpyoKnig/n8n-sandbox-service). The chart change is [PR #1](https://github.com/TpyoKnig/n8n-sandbox-service/pull/1) and the Talos verification is written up in [PR #2](https://github.com/TpyoKnig/n8n-sandbox-service/pull/2).

Sketching the shape here as well, because the shape is what makes it safe to consider, not the fact that it works.

**The runner template became generic over the active runner.** `sysbox-runner-statefulset.yaml` was already the whole in-cluster data plane, so rather than copy it we parameterised it: a small set of helpers resolves the active mode to its values block, component label and resource name, and the StatefulSet renders from those. One template, two modes, no duplicated manifest to drift.

**Each mode keeps its own resource names** — `sysbox-runner` and `dind-runner` rather than a shared `runner`. That is deliberate and not cosmetic. StatefulSet selectors are immutable, so folding both onto one label would rewrite the selector of every existing release and break it on upgrade. The two mode-specific names cost a little repetition and make the upgrade a no-op for current users.

**The dind path is a small delta.** In the template: `privileged: true` in the container security context, and an `emptyDir` at `/var/lib/docker` with a size limit so the inner daemon's layers are bounded. In the values: `runtimeClassName` empty, `hostUsers` unset, and the `sysbox-install` nodeSelector and `sysbox-runtime` toleration cleared, since the point of the mode is that any node can run it.

**Validation refuses the dangerous middles**, since each of them installs cleanly and then fails somewhere far from the cause:

- `dind` without `privileged` — dockerd cannot create the namespaces a sandbox needs, so the runner starts and then fails every sandbox it is asked for
- `dind` with a `runtimeClassName` — that is what sysbox mode is for, and naming one on a cluster that lacks it puts you straight back into the admission rejection above
- `sysbox` with `privileged` — grants the container exactly what sysbox exists to withhold

The existing listener-versus-service port check was generalised to whichever runner is active rather than being sysbox-only.

Two values that were sysbox-specific became mode-neutral (`networkPolicy.sysboxRunner` and `monitoring.serviceMonitor.sysboxRunner`, both now `.runner`). The old keys fail the render with a message naming the new one, rather than being silently ignored, since a silently dropped NetworkPolicy is the kind of thing nobody notices.

**The namespace needs Pod Security Admission at `privileged`.** Worth stating in any docs, because a PSA denial also surfaces as an event on the StatefulSet rather than a failing pod, which is the same misplaced-signal problem described above.

**Regression check.** We rendered the sysbox path against upstream and diffed it across the combinations we cared about: defaults, cert-manager TLS, runner persistence, API ingress, NetworkPolicy peers, `sysboxRunner.enabled=false` and `external` mode. Every rendered Kubernetes object is identical, field for field. The only differences anywhere in the output are the `helm.sh/chart` version label and two `# Source:` comments, because the runner templates were renamed to drop `sysbox-` from the filename. The existing mode is untouched, which was the constraint we set ourselves before starting.

It has been running on Talos Linux v1.13.7 (kernel `6.18.39-talos`) with Kubernetes v1.36.3 and containerd 2.2.6, executing sandboxed code end to end through the n8n AI Assistant, with the inner container correctly reporting the Talos kernel.

## What would help

Roughly in order of how much we would prefer it:

1. A supported in-cluster mode that does not require modifying the node, with the security trade-off documented rather than hidden. Gated behind an explicit acknowledgement value so nobody arrives at it by accident.
2. A statement of direction, if the answer is that Firecracker will be the supported path for these environments. Knowing that is worth as much as the feature, because it tells us whether to invest in a workaround or wait.
3. Failing both, an explicit note in the README that immutable-rootfs distributions cannot run the in-cluster data plane, plus the install-time hint above. Being told clearly up front is much better than finding the reason in a StatefulSet event.

## On contributing it

We are happy to open a PR, and we would rather be told what to build than send the fork as-is.

The fork solved our problem, and our constraints are not necessarily yours. We optimised for not disturbing existing sysbox users, because that was the risk we could see; you can see risks we cannot, and you own the direction. Several of the choices above are ones we would expect a maintainer to have opinions about:

- Whether a third `dataPlane.mode` is the right seam at all, or whether this belongs as a property of the existing in-cluster mode, something like a privilege source, leaving `mode` to describe where the data plane lives rather than how it is privileged.
- Whether mode-specific resource names are worth the duplication, given we chose them for selector immutability on upgrade. If you would accept a breaking change for a cleaner layout, that decision is yours to make and it simplifies the templates.
- How the opt-in should read. We gated it on validation that refuses the unsafe combinations, but an explicit acknowledgement value, a documented warning, or a stricter posture are all reasonable and we have no attachment to ours.
- Whether you want it charted at all, if Firecracker is the intended answer for these environments.
- What the supported matrix should say, since we can only speak for Talos. We have not tested Bottlerocket, Flatcar or Fedora CoreOS, and would not want docs to imply we had.

So: if a contribution is welcome, we would appreciate advisement on the implementation plan you want before we write it. Tell us the seam, the naming, and how explicit the opt-in should be, and we will build to that specification, carry the docs and tests, and verify it against a real Talos cluster since we have one on hand. If the answer is that this does not belong in the chart, that is a useful answer too, and the note in option 3 above would still help the next person who tries.

tomi:
## Summary

Immutable-rootfs distros such as Talos cannot install sysbox, so they had no in-cluster data plane. This PR adds a `privileged` isolation mode and restructures the chart values around it.

- Adds `runner.isolation: privileged` — same Docker-in-Docker runner with `privileged: true`, no sysbox needed.
- Gated behind `runner.acknowledgePrivileged=true`. The render fails without it.
- `dataPlane.mode` is now `in-cluster` or `external` (was `sysbox`).
- One `runner` block replaces `sysboxRunner`, with `runner.sysbox.*` / `runner.privileged.*` for runtime and scheduling.
- Old values fail the render with messages that name the new keys.
- Resource names follow the isolation, so existing sysbox releases upgrade with no StatefulSet changes.
- Chart bumped to `0.4.0`.

Verified with `helm template`: the sysbox render is byte-identical to 0.3.0 except `# Source:` comments and the chart version label. CI now renders both isolations.

Tested on dev cluster in namespace `tomi-privileged-sandbox-test`

Closes #122

## Review / Merge checklist

- [ ] Tests included.


charts/n8n-sandbox-service/templates/cert-manager.yaml
```diff
@@ -53,20 +53,22 @@ spec:
...
   labels:
     {{- include "n8n-sandbox-service.labels" . | nindent 4 }}
-    app.kubernetes.io/component: sysbox-runner
+    app.kubernetes.io/component: {{ $runnerComponent }}
 spec:
   secretName: {{ $runnerRegistration.secretName }}
   duration: {{ $certManager.duration }}
   renewBefore: {{ $certManager.renewBefore }}
-  commonName: {{ include "n8n-sandbox-service.sysboxRunnerName" . }}-registration-client
+  commonName: {{ $runnerName }}-registration-client
```
TpyoKnig: This overruns the 64 byte x509 CommonName limit and cert-manager's webhook rejects the install. `$runnerName` is already capped at 63 for the k8s label limit, then this appends 20 more bytes. `-privileged-runner` is 4 bytes longer than `-sysbox-runner`, so it trips here first: release name `sandbox` renders 65 and fails, while `n8n-sandbox-service` renders 57 and is fine. Same shape `_helpers.tpl` already uses for `runnerName`, so the suffix stays whole and the sysbox render is unchanged.

```suggestion
  {{- $runnerRegistrationCn := "-registration-client" }}
  commonName: {{ printf "%s%s" ($runnerName | trunc (int (sub 64 (len $runnerRegistrationCn))) | trimSuffix "-") $runnerRegistrationCn }}
```

TpyoKnig:
Ran it on my Talos cluster, it works. I got n8n's AI Assistant to build a workflow and execute it through the privileged runner.

My box: Talos v1.13.7, kernel 6.18.39-talos, k8s 1.36.3, containerd 2.2.6, 5 nodes, no sysbox-runc RuntimeClass anywhere. I ran chart 0.4.0 at 7884af0 with your config verbatim, `tls.mode: certManager` against a namespace-local self-signed CA.

I hit all three gates on purpose and they fire with the right message, including the one that catches the `dataPlane.mode: dind` spelling my fork used, so the migration points you at the new key. Runner registered over mTLS at `capacity_total=1000`, sandbox image pulled, and create/exec/delete all behave:

```
{"seq":1,"type":"stdout","data":"Linux-6.18.39-talos-x86_64-with-glibc2.36\n"}
{"seq":3,"type":"exit","exit_code":0,"success":true,"execution_time_ms":29}
```

Then I pointed a fresh n8n 2.34.6 at it and drove the AI Assistant from the browser myself. I asked it for raw sandbox output and it came back with 48611 for the 5000th prime (correct) and `Linux sandbox 6.18.39-talos ... x86_64`, so that ran on my node and nowhere else. Then I had it build, run and save a three node workflow, Code node returning `{"date":"2026-08-19","randomNumber":825499}`. 346 requests over the session: 155 `mkdir`, 151 `PUT /files` as it uploads the template knowledge base, 8 executions. All 2xx except four 404s, which are n8n's own first-run probes for files it hasn't written yet. `NOTES.txt` did its job on me too, it printed the StatefulSet name and the PSA label line before I could go hunting for pods that don't exist.

## One bug, and it blocks a plain helm install

With `tls.mode: certManager`:

```
Error: INSTALLATION FAILED: server-side apply failed for object
n8n-sbx-126/sandbox-n8n-sandbox-service-privileged-runner-registration-client
cert-manager.io/v1, Kind=Certificate: admission webhook "webhook.cert-manager.io"
denied the request: spec.commonName: Too long: must have at most 64 bytes
```

`runnerName` is capped at 63 for the k8s label limit, then cert-manager.yaml appends a 20 byte suffix to build `commonName`, and x509 caps CN at 64. Two lines, L47 on the API control client and L71 on the runner.

Not new in #126, but #126 moves the threshold. `-privileged-runner` is four bytes longer than `-sysbox-runner`, so for a release name that doesn't already contain the chart name the ceiling drops from 10 characters to 6:

| Release name | sysbox CN | privileged CN |
| --- | --- | --- |
| `n8n-sandbox-service` (quickstart) | 53 | 57 |
| `sandbox` | 61 | 65, fails |

So your own quickstart is safe and a shorter, more natural release name isn't. Anyone hitting this on an upgrade from 0.3.x will blame the isolation switch, correctly.

I've left a suggestion on L71 so you can take it with one click. It uses the same `sub (len $suffix)` shape `_helpers.tpl` already uses for `runnerName`, so the base truncates and the suffix stays whole. Sysbox render is unchanged, so your byte-identical upgrade guarantee holds. I verified it by applying the rendered Certificates straight at my webhook: before, rejected, after, all four created. CN isn't read anywhere in the Go code so truncating costs nothing.

L47 is the same bug on the API control client, but #126 doesn't touch that line so I can't suggest on it. It only bites at a longer release name than the runner one does, and it's pre-existing, so it's your call whether it rides along here or gets its own issue. Happy to file it if you'd rather keep this PR scoped.

## hostUsers: false doesn't work on Talos

This is the doc thing you asked about. The caveat reads like the version floor is the constraint. On Talos you can clear it and still be stuck. Mine clears 1.33/2.0 comfortably at 1.36.3/2.2.6 and still can't use it:

```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create
network namespace for sandbox "...": failed to start noop process for unshare:
fork/exec /proc/self/exe: no space left on device
```

Talos ships `user.max_user_namespaces=0` so the kernel refuses the unshare. It surfaces as ENOSPC and has nothing to do with disk, same error that kills rootless buildkit on Talos, which is where I first met it. An operator can raise the sysctl via `machine.sysctls`, but that's node-level and outside the chart's reach. A sentence naming the sysctl and the misleading ENOSPC would save someone a long detour, "no space left on device" sends you straight to the wrong place.

Worth a line somewhere too: recovery isn't automatic. The wedged pod stays in `ContainerCreating` and the StatefulSet controller won't replace it, needs `kubectl delete pod --force --grace-period=0`. A `helm upgrade --wait` in between leaves the release in `pending-upgrade` and needs a rollback.

One smaller thing I tripped over. The registration retries read as a NetworkPolicy denial: while the API is still starting the runner logs `connect: operation not permitted` three or four times before the backoff succeeds around 15s. Harmless, but "operation not permitted" is exactly what a denied NetworkPolicy looks like and `networkPolicy.enabled` is a thing this chart has, so my first instinct was to go check my policies.

Retiring my fork. #126 covers everything mine did and adds the acknowledgement gate, the long-name collision fix, the bounded emptyDir and the Kata escape hatch. The seam you landed on is better than the one I proposed.

Happy to run anything else you want against my cluster before release. Thoughts?

TpyoKnig:
Correction on one of the smaller notes above: scratch the `appVersion` one.

I said the chart reads as pinning a downgrade because I arrived with the API on 1.2.0. That was self-inflicted. We bumped that tag by hand to force a re-pull off a stale `:latest`, so the gap was mine and I read it straight back as yours. Nothing inconsistent shipped. I have edited it out of the comment above rather than leave it sitting in the thread as a finding, and apologies if it sent you looking.

The `commonName` bug and the `hostUsers: false` note both stand unchanged.

tomi:
Thank you @TpyoKnig ! Addressed your feedback. Please have a look 🙏

endor (approved):

TpyoKnig:
@tomi  I re-tested `cb897cc9` on Talos v1.13.7 / k8s v1.36.3, fresh namespace, release name `sandbox`, no `fullnameOverride`. That combination is what cert-manager rejected at 65 bytes before, and it installs clean now. Runner registered over mTLS in 25s and code ran on the node:

```
Linux-6.18.39-talos-x86_64-with-glibc2.36
```

The disk quota work holds up under privileged DinD too. With `defaultDiskQuotaMb=512` and `diskQuotaPoolSizeGb=10` on a 20Gi emptyDir:

```
[start-runner] storage pool mounted: /var/lib/docker-pool/docker.img on /var/lib/docker (xfs+prjquota)
[start-runner] disk quota enforcement: ENABLED (pool 10G, per-sandbox 512M)
```

`dd` of 700M stopped at 536805376 bytes, exit 1. All four render guards fired on the configurations they describe.

One thing left, and it is the L47 case I mentioned earlier. It now looks in scope, because it is the last path to the same error this PR set out to remove:

```
commonName: {{ include "n8n-sandbox-service.apiName" . }}-control-client
```

`apiName` caps at 63 and then 15 bytes are appended, so the CN passes 64 once the release name reaches 26 characters. It is isolation-independent, so sysbox users hit it as well.

| Release name length | CN bytes |
| --- | --- |
| 24 | 63 |
| 25 | 64 |
| 26 | 65, rejected |
| 53 | 78, rejected |

Against the live webhook rather than `--dry-run=server`, which does not call it:

```
$ kubectl apply --dry-run=server -f <rendered api control-client Certificate>
admission webhook "webhook.cert-manager.io" denied the request:
spec.commonName: Too long: must have at most 64 bytes
```

`render-tests.sh` will not catch this one. The long-name loop uses `--show-only templates/runner-statefulset.yaml`, so `cert-manager.yaml` never renders inside it.

The pattern you took for the runner registration CN covers it as is. L47 sits outside the diff, so this is a patch rather than a suggestion:

```diff
-  commonName: {{ include "n8n-sandbox-service.apiName" . }}-control-client
+  {{- $apiControlCn := "-control-client" }}
+  commonName: {{ printf "%s%s" ((include "n8n-sandbox-service.apiName" .) | trunc (int (sub 64 (len $apiControlCn))) | trimSuffix "-") $apiControlCn }}
```

If you want the test to cover it, adding a second `--show-only templates/cert-manager.yaml` render inside the existing length loop and asserting every CN is 64 or under catches both certificates.

PR #127, opened 2026-08-19
feat: document security model and write more security related e2e tests

endor:
## Summary

* Document security model
* Add more security-related e2e tests
* Fix rules to drop host-local traffic in sysbox setup

---
## Summary by cubic
Documents the security model and hardens Sysbox network policy to meet CP-2673 tenant isolation. Previously, `DOCKER-USER` only filtered forwarded traffic so sandboxes could reach runner host-local ports; now bridge-level INPUT rules block sandbox-initiated connections to the runner while allowing established traffic.

- Docker runtime: add bridge-wide egress chain in `DOCKER-USER` and a bridge-wide host block in `INPUT`, both matched on the bridge interface; keep per-container ingress chain to allow gateway→daemon only; refuse to apply rules if the bridge iface is missing. Behavior change: runner ports (e.g., 8080, 9091) are unreachable from sandboxes.
- Runner: detect bridge iface via network inspect (use `com.docker.network.bridge.name` or derive `br-<id>`); on create, set the bridge name option; teardown removes legacy per-container jumps from `DOCKER-USER` and `INPUT`.
- API tests: new `internal/api/handlers_proxy_tenant_test.go` proves tenant checks prevent proxying (404) and no request reaches a runner.
- E2E: extend isolation to two tenants with a positive control, assert guests cannot reach the runner via their gateway, expand blocked CIDRs (100.64.0.0/10, 198.18.0.0/15, 240.0.0.0/4), and verify interface-based policy cannot be bypassed by adding an IP; helpers export `BASE_URL` and use `@n8n/sandbox-client`.
- Docs: add `docs/security-model.md` and link it from `README.md` and `docs/architecture.md`, detailing trust boundaries and non-guarantees.

<sup>Written for commit 954a31fce5dca880ba2c641f7c56c3c32f91ec76. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,7 @@ Registration vs lifecycle: Runners dial the API over gRPC for registration. Sand
 
 Debugging gRPC: See [`docs/grpcurl-debug.md`](docs/grpcurl-debug.md).
 
-Security FAQ (draft): See [`docs/security-faq.md`](docs/security-faq.md).
-
-Weak points + hardening plan (draft): See [`docs/security-weak-points-and-hardening.md`](docs/security-weak-points-and-hardening.md).
+Trust boundaries and non-guarantees: See [`docs/security-model.md`](docs/security-model.md).
 
 Bearer token: Still required in metadata (`Authorization: Bearer …`) in addition to mTLS for registration.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,6 +16,7 @@ The n8n Sandbox Service provides isolated execution environments via a REST API.
 ## Reference
 
 - [Architecture](architecture.md) — system design, component overview, communication patterns, and deployment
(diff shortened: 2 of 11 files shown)


tomi (commented):
LGTM. There are couple cubic findings worth checking

PR #129, opened 2026-08-20
fix: stop counting failed requests as sandbox activity (CP-2670)

endor:
## Summary

The proxy stamped `last_active_at` and reset the status to running before forwarding the request, so a sandbox that failed every call still looked freshly used. Clients retry failures, which kept broken sandboxes holding a slot longer than necessary. The write now happens once the sandbox has answered, and is skipped on 5xx and on an unreachable runner. A 4xx still counts, since the sandbox was up and rejected the request on its merits.

---
## Summary by cubic
Stops counting failed proxy requests as sandbox activity so the idle sweeper can reclaim broken sandboxes instead of client retries keeping them alive. Previously the proxy set last_active_at and status=running before forwarding; now it updates only after a non-5xx upstream response, skips unreachable runners, still counts 4xx, and never counts a runner “sandbox gone” 404. Aligns with CP-2670.

- Moves activity stamping into the response hook in `internal/api/handlers.go` via `markSandboxActive`; skips updates on 5xx and connect errors.
- Reaps on runner “sandbox gone” 404 via `reapSandboxIfRunnerGone`; short-circuits activity and still avoids updates even if `Delete` fails.
- Adds tests: `internal/api/handlers_proxy_activity_test.go` (5xx/unreachable no-activity; 4xx counts) and extends `internal/api/handlers_reap_test.go` to cover failed-delete reaps.
- Updates `docs/API.md` to document activity semantics (502/503 don’t count; 4xx do, except runner 404) and clarify exec streams that end without a terminal event.

<sup>Written for commit 9a6fb67a207905576c842eb9dd9eb04b2671d429. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/129?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


internal/api/handlers.go
```diff
@@ -481,13 +481,33 @@ func newRunnerReverseProxy(runnerURL *url.URL, runnerAPIKey string, onResponse f
 	}
 }
 
-func reapSandboxIfRunnerGone(s store.SandboxStore, sandboxID string, resp *http.Response) {
+// reapSandboxIfRunnerGone drops the record when the runner no longer knows the
+// sandbox, and reports whether the runner said so. A failed delete still
```
tomi: This is quite ambiguous what this returns. Can make it clearer?

tomi (approved):
LGTM, one nit about a comment clarity

internal/api/handlers.go
endor: Fixed.

PR #149, opened 2026-09-01
fix: keep an execution delete from spending the crash restart report

tomi:
Fixes [CAT-4310](https://linear.app/n8n/issue/CAT-4310/docker-crash-recovery-e2e-flakes-an-sdk-cleanup-delete-can).

## The contract

A `DELETE` of an execution must **not** consume the restart report.

The runner reports a crash once, with `409 sandbox_restarted`, to the request
that wakes the sandbox. That report is meant for a client. A `DELETE` of an
execution carries no client intent: the SDK sends one in the background after
every command (`sdk/src/exec.ts:82`) and discards the answer.

So this was not only a test flake. In production, a crash that landed in that
window was reported to a request nobody reads, and the client's next `exec` got
a healthy-looking `200` on a sandbox whose processes were gone — the exact
failure the `409` exists to prevent.

## The fix

`DELETE /sandboxes/{id}/executions/{exec_id}` now answers `204` without waking a
sandbox that is not running.

The `204` is honest rather than a dodge that keeps the report. An execution
lives only in the guest's memory, so a sandbox that is stopped, or that came
back from a crash, has already done what the delete asks for. The `409` waits
for the next request a client reads.

The handler keys off `ErrSandboxNotRunning`, so it covers both backends: the
Docker restart mark and a Firecracker sandbox torn down by its crash.

## Tests

`TestDeleteExecutionDoesNotWakeOrSpendTheRestartReport` — the delete gets `204`,
drives no wake, reaches no daemon, and a following exec still gets its `409`.
`TestDeleteExecutionIsProxiedWhenTheSandboxIsRunning` covers the ordinary path.

## Docs and comments

- `docs/API.md` — the endpoint documents that it never wakes and never returns
  `409`. The crash-loss list says a delete returns `204`.
- `docs/architecture.md` — why a request with no client behind it must not spend
  the report.
- `sandbox-crash-recovery-docker.spec.ts` — `waitForDeath` claimed the next
  request is *guaranteed* to meet the `409`. That was never true for a request
  already in flight, which is what made the flake invisible. It now says what it
  guarantees, and names the runner behaviour the test relies on.
- `helpers.ts` — `crashGuest` said both SDK follow-up routes wake a stopped
  sandbox. Only the `GET` does now.

The spec itself needed no change, and the SDK needed none either.

## Checks

`make fmt-check`, `make vet`, `make test` pass. The e2e spec needs Docker and a
live runner, so it was not run locally — the new unit test covers the race
deterministically.

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/runner/proxy.go
+++ b/internal/runner/proxy.go
@@ -24,15 +24,33 @@ type proxyTarget struct {
 
 // ProxyHandler returns a handler that reverse-proxies requests to the sandbox daemon.
 func ProxyHandler(rt runnerruntime.Runtime, cfg *config.Config, rec *metrics.RunnerRecorder) http.HandlerFunc {
-	return proxyHandler(rt, cfg, rec, false)
+	return proxyHandler(rt, cfg, rec, false, true)
 }
 
 // UploadProxyHandler is like ProxyHandler but enforces cfg.MaxFileBytes on the request body.
 func UploadProxyHandler(rt runnerruntime.Runtime, cfg *config.Config, rec *metrics.RunnerRecorder) http.HandlerFunc {
-	return proxyHandler(rt, cfg, rec, true)
+	return proxyHandler(rt, cfg, rec, true, true)
 }
 
-func proxyHandler(rt runnerruntime.Runtime, cfg *config.Config, rec *metrics.RunnerRecorder, limitBody bool) http.HandlerFunc {
+// DeleteExecutionHandler serves DELETE /sandboxes/{id}/executions/{exec_id}, the
+// one sandbox route that answers without waking.
+//
+// An execution lives only in the guest's memory, so a sandbox that is stopped, or
+// that came back from a crash, has already lost the one this names and the delete
+// has nothing left to do. It gets 204.
+//
+// Not waking is what keeps the crash report honest. A recovery reports the restart
+// to the single request that wakes the sandbox, and this delete is not a client
+// asking to use it: the SDK sends one in the background after every command and
+// discards the answer. Letting it spend the report would hide the crash from the
+// request that comes next.
+func DeleteExecutionHandler(rt runnerruntime.Runtime, cfg *config.Config, rec *metrics.RunnerRecorder) http.HandlerFunc {
+	return proxyHandler(rt, cfg, rec, false, false)
+}
+
+// wake says whether a sandbox that is not running is started for the request. Only
+// DeleteExecutionHandler passes false; see there for why.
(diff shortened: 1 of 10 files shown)


tomi:
The first push still failed CI, for a reason worth recording.

My handler checked whether the sandbox was running and then handed the request to
the proxy, which checked again. A daemon lookup is a `docker ps` plus an `inspect`,
so under CI load the two checks sit ~150ms apart — wide enough for a whole crash.

From the failed run:

```
59.134  DELETE /executions/82969eb6   arrives, sandbox alive, first lookup says running
59.156  POST   /executions            crashGuest: kill -TERM 1
59.290  WARN   docker guest died      the restart mark is set
59.722  INFO   docker sandbox recovered
59.722  DELETE /executions/82969eb6   409 after 588ms: the proxy's own lookup drove the wake
59.828  POST   /executions            the test's request: 200
```

So the fix reproduced the very race it was meant to close, one layer up.

The wake is now a parameter of the single lookup `resolveDaemonURL` already does, so
the delete route decides once and a later crash cannot change an answer already
given. The test gained the case that catches it: a runtime whose sandbox dies on the
*second* lookup, asserting the handler never makes one.

internal/runner/proxy.go
```diff
@@ -24,15 +24,33 @@ type proxyTarget struct {
...
 func UploadProxyHandler(rt runnerruntime.Runtime, cfg *config.Config, rec *metrics.RunnerRecorder) http.HandlerFunc {
-	return proxyHandler(rt, cfg, rec, true)
+	return proxyHandler(rt, cfg, rec, true, true)
 }
 
-func proxyHandler(rt runnerruntime.Runtime, cfg *config.Config, rec *metrics.RunnerRecorder, limitBody bool) http.HandlerFunc {
+// DeleteExecutionHandler serves DELETE /sandboxes/{id}/executions/{exec_id}, the
+// one sandbox route that answers without waking.
+//
+// An execution lives only in the guest's memory, so a sandbox that is stopped, or
```
endor, line 38: We have this same documentation in at least three places in this PR. I know we always do this, but maybe we should try to slim our documentation down a bit.

endor (approved):

PR #154, opened 2026-09-02
fix(chart): bound the sysbox data root and refuse a PVC in a user namespace

tomi:
Two problems in the same place, found while debugging a sysbox runner that failed to start after an upgrade.

## Problem 1: the sysbox data root was unbounded

`runner-statefulset.yaml` withheld the data root volume from sysbox isolation:

```gotemplate
{{- $dockerEmptyDir := and $privileged (not $runner.dockerDataRoot.persistence.enabled) }}
```

So a sysbox runner without persistence — the chart default — had **no volume at all** at `/var/lib/docker`. Image layers and sandbox filesystems went to the container filesystem with no size limit, where they can fill the node disk. Privileged isolation got a bounded `emptyDir` for exactly that reason; sysbox did not.

## Problem 2: persistence breaks inside a user namespace

The reverse case fails outright. `persistence.enabled` with `sysbox.runtime.hostUsers: false` mounts a PersistentVolume at `/var/lib/docker` inside a user namespace. dockerd chmods its data root at startup and does not own the volume root, so it exits immediately:

```
level=info msg="Starting up"
level=info msg="Daemon shutdown complete" error="chmod /var/lib/docker: operation not permitted"
```

The runner never becomes ready. The pod UID range can also change when the pod is recreated, so a runner that works today can fail on the next rollout.

This is a configuration the docs recommend. `render-tests.sh` even covered "the README example with persistence", but with `defaultDiskQuotaMb=2048`, which moves the volume to `/var/lib/docker-pool` and misses the broken path. Sysbox + persistence + no quotas had no test.

## Changes

**1. Both isolations always hold a data root volume.**

```diff
-{{- $dockerEmptyDir := and $privileged (not $runner.dockerDataRoot.persistence.enabled) }}
+{{- $dockerEmptyDir := not $runner.dockerDataRoot.persistence.enabled }}
```

The claim when persistence is on, the bounded `emptyDir` otherwise. Three knock-ons followed, and the render tests caught two of them:

- `configmap.yaml` duplicated the old condition and withheld `SANDBOX_RUNNER_DISK_QUOTA_POOL_PATH` from the same configurations.
- The statefulset volume mount condition became `or X (not X)`, so it is now unconditional.
- The quota guard that failed the render when no volume held the pool image can no longer happen, so it and its two tests go. Sysbox with quotas and no persistence now renders, with the pool on the bounded `emptyDir`.

**2. The broken combination fails the render.**

`validation.yaml` refuses sysbox + `hostUsers: false` + persistence, unless quotas are on. The message names the dockerd error so an operator who already hit it can search for it. Quotas are exempt because they give dockerd a freshly made xfs image as its data root, which the container owns.

`runner.dockerDataRoot.persistence.acknowledgeUserNamespace: true` keeps the combination available, following the existing `runner.acknowledgePrivileged` idiom. A CSI driver that id-maps volumes handles this correctly and the chart cannot tell, so an unconditional failure would break clusters that work today.

**3. Docs.** A new README section, `Docker Data Root`, covers which volume you get and why persistence and user namespaces do not mix. It also records something worth knowing: persistence buys a warm image cache and nothing else. It does not carry sandboxes across a restart, because `reconcileContainers` deletes every sandbox container the runner finds at startup and again at shutdown. `quickstart-k8s.md` gets a troubleshooting entry for the runtime symptom.

I also moved the `emptyDir` paragraph out of the Privileged Isolation section, where it was correct but now applies to both isolations.

## Verification

`charts/n8n-sandbox-service/render-tests.sh` passes, with new cases for:

- the default sysbox render mounting `emptyDir` at `/var/lib/docker` with its `sizeLimit`
- persistence replacing that with a claim, and no `sizeLimit` left behind
- the userns combination failing, the acknowledgement letting it through, and `hostUsers: null` not tripping the guard
- privileged isolation being unaffected
- sysbox with quotas and no persistence landing the pool on the bounded volume

`helm lint` is clean. `shfmt` and `shellcheck` are clean on `render-tests.sh`, with the CI-pinned 3.13.1 and 0.11.0. Chart version bumped 0.6.2 → 0.6.3. No Go files changed.

## What this does not fix, and what I could not verify

**Persistence stays unusable in a user namespace** without the acknowledgement. The real fix needs a pod `fsGroup` plus a data root in a subdirectory the container creates, so dockerd chmods a directory it owns. That belongs in its own PR because it silently migrates existing data roots, and because whether the kubelet applies `fsGroup` to a CSI PVC in a userns pod depends on the driver and Kubernetes version.

**Change 1 is a runtime behaviour change I could only verify by rendering.** Mounting an `emptyDir` at `/var/lib/docker` for sysbox is new. Sysbox may currently back that path with a host directory it manages, so this moves where the data lives and brings it under the pod's ephemeral-storage accounting — which is the intent, but it needs one test on a sysbox node before merge. I have no sysbox cluster.

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/charts/n8n-sandbox-service/README.md
+++ b/charts/n8n-sandbox-service/README.md
@@ -64,7 +64,7 @@ Requirements and recommendations:
 - Prefer a dedicated node pool for the runner via `runner.privileged.scheduling`.
 - Platforms that block privileged containers outright (for example GKE Autopilot) cannot use this isolation; use `dataPlane.mode: external` there.
 
-When `runner.dockerDataRoot.persistence` is disabled, the chart mounts an `emptyDir` with `runner.dockerDataRoot.emptyDir.sizeLimit` at `/var/lib/docker`, so the inner Docker daemon's layers are bounded. With disk quotas the volume holds the quota pool image instead; see [Disk Quotas](#disk-quotas).
+The chart bounds the inner Docker daemon's data root for both isolations; see [Docker Data Root](#docker-data-root).
 
 Two hardening options:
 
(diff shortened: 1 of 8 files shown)


charts/n8n-sandbox-service/templates/validation.yaml
```diff
@@ -48,16 +48,24 @@
 {{- fail "runner.config.controlGrpcListenAddr port must match runner.service.controlGrpcPort" -}}
 {{- end -}}
 {{- end -}}
-{{- if and (eq .Values.dataPlane.mode "in-cluster") (gt (int .Values.runner.config.defaultDiskQuotaMb) 0) -}}
+{{- if eq .Values.dataPlane.mode "in-cluster" -}}
 {{- $dockerDataRoot := .Values.runner.dockerDataRoot -}}
-{{- $chartVolume := or $dockerDataRoot.persistence.enabled (eq .Values.runner.isolation "privileged") -}}
-{{- if and (not $chartVolume) (not .Values.runner.config.diskQuotaPoolPath) -}}
-{{- fail "runner.config.defaultDiskQuotaMb above 0 needs a bounded volume for the disk-quota pool image, and this configuration has none. The image would land on the container filesystem and could fill the node disk. Set runner.dockerDataRoot.persistence.enabled=true, or set runner.config.diskQuotaPoolPath to a path on a volume you supply through runner.extraVolumes. See the chart README section 'Disk Quotas'" -}}
+{{- $sysboxUserns := and (eq .Values.runner.isolation "sysbox") (kindIs "bool" .Values.runner.sysbox.runtime.hostUsers) (not .Values.runner.sysbox.runtime.hostUsers) -}}
```
endor, line 53: This whole code is pretty hard to read. I wonder whether there's a way to write this more readably. The table in the README helps a bit, but it's hard to do the translation from one to the other.

charts/n8n-sandbox-service/README.md
```diff
@@ -144,6 +144,37 @@ runner:
...
+
+```
+level=info msg="Daemon shutdown complete" error="chmod /var/lib/docker: operation not permitted"
+```
+
+The runner then never becomes ready. The pod's UID range can also change when the pod is recreated, so a runner that works today can fail on the next rollout.
+
+The chart fails the render for `runner.isolation: sysbox` with `runner.sysbox.runtime.hostUsers: false` and persistence enabled, unless disk quotas are on. Quotas move the volume to `/var/lib/docker-pool` and give the daemon a freshly made xfs image as its data root, which the container owns.
+
+Three ways out, best first:
```
endor, line 172: Instead of describing a problem and "three ways out", maybe we should just provide different safe ways to configure this chart.

charts/n8n-sandbox-service/values.yaml
```diff
@@ -207,15 +207,27 @@ runner:
...
   dockerDataRoot:
     persistence:
       enabled: false
       size: 64Gi
       storageClass: ""
       accessModes:
         - ReadWriteOnce
-    # Used only with isolation=privileged when persistence is disabled.
-    # Bounds the inner Docker daemon's image layers and sandbox filesystems.
+      # A PersistentVolume at /var/lib/docker breaks the inner Docker daemon
```
endor, line 221: We're repeating the same thing a couple of times.

PR #155, opened 2026-09-03
feat: ephemeral sandboxes

endor:
## Summary by cubic
Adds ephemeral sandboxes that are deleted instead of stopped when idle. Previously an idle sandbox stopped and could be woken; a sandbox created with `ephemeral: true` is now deleted at the idle-stop window (or the idle-delete window when idle stop is disabled), never reports `stopped`, and returns 404 on every request after that window.

- The API accepts `ephemeral` on create and returns it in every sandbox response; the flag is fixed at creation, so reconnecting to an existing id keeps the original value.
- The idle sweeper deletes ephemeral sandboxes where it would stop a regular one; the safety buffer now also applies when only the idle-stop window is configured, and ephemeral sandboxes are never reclaimed when both idle windows are disabled.
- The 404 fence takes precedence over routing validation, so an expired sandbox returns 404 even when its row lacks runner routing information.
- The SDK adds `createSandbox({ ephemeral: true })`; a create with the flag but no caller-supplied id is no longer marked retry-safe, since retrying would provision a second sandbox.

**Migration**

- Adds an `ephemeral` column to the `sandboxes` table in both SQLite and Postgres; existing rows default to `false`, so no backfill is needed.

<sup>Written for commit 122a8d527984251510510c1960f4fcca9d1c411f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


endor:
@cubic re-review this PR and resolve comments that have been dealt with.

tomi (commented):
Couple comments, all minor

internal/api/ttl.go
```diff
@@ -151,10 +171,40 @@ func withLockedSandbox(ctx context.Context, s store.SandboxStore, id string, fn
...
 
+// deleteIdleSandbox deletes rec on the runner, then in the store. Callers hold
+// the sandbox lock. Any failure leaves the row in place so the next sweep retries.
+func deleteIdleSandbox(ctx context.Context, s store.SandboxStore, reg registry.RunnerRegistry, cfg *config.APIConfig, tlsCfg *runnerctl.TLS, rec *store.SandboxRecord, now time.Time, reason string) {
+	if orphanReapDue(reg, rec.RunnerID, cfg, now) {
+		reapOrphanSandbox(s, rec, rec.RunnerID)
+		return
+	}
+	controlAddr := resolveControlAddr(rec, reg)
+	if err := runnerctl.DeleteSandbox(ctx, controlAddr, cfg.RunnerAPIKey, tlsCfg, rec.ID); err != nil {
```
tomi, line 182: Started wondering, do we handle a case where the runner crashes and all sandboxes are lost? Or do we try to retry the deletion forever? Nothing to do with this PR, but perhaps needs a followup unless we handle it already?

internal/api/ttl.go
```diff
@@ -171,25 +221,43 @@ func sweepIdleDeleteSandboxes(ctx context.Context, s store.SandboxStore, reg reg
...
+		return
+	}
+
+	for _, rec := range records {
+		if rec == nil || !rec.Ephemeral {
+			continue
+		}
+		id := rec.ID
+		err := withLockedSandbox(ctx, s, id, func(rec *store.SandboxRecord) {
+			if !rec.Ephemeral || rec.Status != "running" || rec.LastActiveAt > cutoff {
```
tomi: Do we need to check Ephemeral twice? We have an invariant that it is immutable, right?

internal/api/ttl.go
```diff
@@ -213,7 +280,8 @@ func sweepIdleStopSandboxes(ctx context.Context, s store.SandboxStore, reg regis
 		}
 		id := rec.ID
 		err := withLockedSandbox(ctx, s, id, func(rec *store.SandboxRecord) {
-			if rec.Status != "running" || rec.LastActiveAt > stopCutoff {
+			// Ephemeral rows are deleted by sweepEphemeralSandboxes, never stopped.
+			if rec.Status != "running" || rec.Ephemeral || rec.LastActiveAt > stopCutoff {
```
tomi: We could include the `ephemeral` as a where condition into `ListForIdleReapStop` so we don't get the rows we are not interested in

sdk/src/sandboxes.ts
```diff
@@ -6,11 +6,17 @@ export async function createSandbox(
   http: HttpClient,
   options?: CreateSandboxOptions,
 ): Promise<SandboxRecord> {
+  const data: Record<string, unknown> = {};
+  if (options?.id !== undefined) data.id = options.id;
+  if (options?.ephemeral !== undefined) data.ephemeral = options.ephemeral;
+
   const response =
-    options?.id !== undefined
+    Object.keys(data).length > 0
```
tomi: Does the create endpoint accept an empty object? We could simplify this by removing the condition if it does

docs/API.md
```diff
@@ -158,6 +159,17 @@ Create a sandbox. With no request body, the service generates a UUID. Callers ma
 
 If that ID still belongs to the caller and is within its idle-delete window, the existing sandbox is returned. If it has passed the window, the stale sandbox is deleted before the ID is reused. If the ID belongs to another tenant (or to admin when the caller is a tenant), the request fails with `409`.
 
+**Request body fields** (all optional):
+
+- `id` — lowercase UUID to create or reconnect to, as above.
+- `ephemeral` (boolean, default `false`) — delete the sandbox, instead of stopping it, once it has been idle for `SANDBOX_API_IDLE_STOP_AFTER` (default `1h`; `SANDBOX_API_IDLE_DELETE_AFTER` when idle stop is disabled). It never reports `status: "stopped"`; once the window has passed, `GET`, exec and file requests return `404` and the sweeper removes it. Fixed at creation: reconnecting to an existing `id` keeps the value it was created with.
```
tomi: I don't think we need to repeat the logic behind this everywhere. IMO it's fine to say delete the sandbox, instead of stopping it when the idle window has passed.

sdk/README.md
```diff
@@ -43,6 +43,11 @@ const stableSandbox = await client.createSandbox({
   id: '550e8400-e29b-41d4-a716-446655440000',
 });
 
+// Ephemeral: deleted instead of stopped once idle past the service's
```
tomi: IMO we don't need every example here either. IMHO we should clean this file

endor: This is the only documentation for the SDK we have at the moment, right? I would maybe leave it in until we have a better way of documenting.

internal/api/ttl.go
endor: It's immutable, yep.

internal/api/ttl.go
endor: Yeah, we handle this case. 

1. Runner crashes and comes back: sweeper calls `DeleteSandbox` on the runner, and the runner treats "unknown sandbox" as success, API's delete succeeds, the store row is removed, no infinite retry.
2. Runner never comes back, or comes back under a new ID: `deleteIdleSandbox` first checks `GoneLongEnough(runnerID, 5m)`. If true, `reapOrphanSandbox` deletes the store row without contacting the runner.
3. Runner is registered but returns any other error (Unavailable, deadline): retry every sweep, no backoff, no counter. That's fine as long as (1) or (2) eventually applies.

There's an edge case which is covered in CP-3177.

internal/api/ttl.go
endor: Fixed.

sdk/src/sandboxes.ts
endor: Fixed.

docs/API.md
endor: Tried to reduce the duplication in documentation.

sdk/README.md
tomi: Fair enough. Let's keep it 👍

internal/api/ttl.go
```diff
@@ -238,20 +238,19 @@ func sweepIdleDeleteSandboxes(ctx context.Context, s store.SandboxStore, reg reg
 func sweepEphemeralSandboxes(ctx context.Context, s store.SandboxStore, reg registry.RunnerRegistry, cfg *config.APIConfig, tlsCfg *runnerctl.TLS, now time.Time) {
 	cutoff := now.Unix() - idleSeconds(ephemeralIdleWindow(cfg)) - idleSeconds(cfg.IdleDeleteSafetyBuffer)
 
-	// Running rows idle past cutoff; the ephemeral ones are filtered here.
-	records, err := s.ListForIdleReapStop(cutoff)
+	records, err := s.ListForIdleReapStop(cutoff, true)
```
tomi, line 241: nit: could document what the boolean represents

```go
	records, err := s.ListForIdleReapStop(cutoff, /* ephemeral= */true)
```

internal/api/ttl.go
```diff
@@ -268,7 +267,7 @@ func sweepEphemeralSandboxes(ctx context.Context, s store.SandboxStore, reg regi
 func sweepIdleStopSandboxes(ctx context.Context, s store.SandboxStore, reg registry.RunnerRegistry, cfg *config.APIConfig, tlsCfg *runnerctl.TLS, now time.Time) {
 	stopCutoff := now.Unix() - idleSeconds(cfg.IdleStopAfter)
 
-	records, err := s.ListForIdleReapStop(stopCutoff)
+	records, err := s.ListForIdleReapStop(stopCutoff, false)
```
tomi, line 270: nit: could document what the boolean represents

```go
	records, err := s.ListForIdleReapStop(cutoff, /* ephemeral= */false)
```

tomi (approved):
LGTM, couple nits, not blocking

PR #173, opened 2026-09-09
fix: update example compose files for tls

### LINKED ISSUE #166: docs/examples/compose.linux.yaml: SANDBOX_RUNNER_HTTP_BASE_URL host is not a SAN of the runner certificate
### Summary

In `docs/examples/compose.linux.yaml` the runner advertises an HTTP base URL whose host is not a SAN of the certificate the runner serves on `:8080`. Every API → runner HTTP call (exec, file transfer) fails the TLS handshake: sandboxes are created and immediately torn down, and the caller gets `runner unavailable`.

### Details

`tls-init` issues the SandboxControl certificate with:

```yaml
--control-san-prefix n8n-sandbox-runner-local
```

so the runner's certificate SAN is `n8n-sandbox-runner-local-1`.

The runner then advertises:

```yaml
SANDBOX_RUNNER_HTTP_BASE_URL: https://n8n-sandbox-runner-1:8080
```

The API uses that URL's host as the TLS server name for HTTP traffic, and `n8n-sandbox-runner-1` is not a SAN of that certificate, so the handshake fails.

For gRPC the same file works around this explicitly:

```yaml
SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_SERVER_NAME: n8n-sandbox-runner-local-1
```

There is no equivalent override for the HTTP path, which is why only HTTP breaks.

Note that the runner's network aliases only include `n8n-sandbox-runner-1`, so changing the base URL host alone is not enough — `n8n-sandbox-runner-local-1` has to resolve as well.

`docs/quickstart-linux.md` already states the constraint under "mTLS certificate details": *"the host in `SANDBOX_RUNNER_HTTP_BASE_URL` must be one of that certificate's SANs"*. The example compose does not satisfy it. It looks like it was introduced by #135 ("feat: encrypt http api <> runner traffic"), which moved the HTTP path to TLS.

### Symptoms

Runner log, repeated on every attempt:

```
http: TLS handshake error from 172.x.x.x:xxxxx: remote error: tls: bad certificate
```

The API creates a sandbox and deletes it right away; the caller sees `runner unavailable`.

### Suggested fix

Two lines in `docs/examples/compose.linux.yaml`:

```yaml
      SANDBOX_RUNNER_HTTP_BASE_URL: https://n8n-sandbox-runner-local-1:8080
```

```yaml
      networks:
        sandbox-service:
          aliases:
            - n8n-sandbox-runner-1
            - n8n-sandbox-runner-local-1
```

Alternatively, add `n8n-sandbox-runner-1` to the SANs of the SandboxControl certificate.

Verified on our deployment: with those two changes the handshake succeeds and `POST /sandboxes` followed by `POST /sandboxes/{id}/executions` works end to end.

### Environment

- sandbox-service 1.3.3
- Debian 13, kernel 6.12, Docker 29.8 / containerd 2.3.4
- sysbox-ce 0.7.1, runner running with `runtime: sysbox-runc` (not privileged)

endor:
## Summary

Fixes #166.

`docs/examples/compose.linux.yaml` and `compose.macos.yaml` issued the runner's SandboxControl certificate for `n8n-sandbox-runner-local-1` while the runner advertised `https://n8n-sandbox-runner-1:8080`. The API verifies HTTP connections against the dialed host, so every exec/file call failed with `tls: bad certificate and the caller saw runner unavailable`. The root `compose.yaml` was unaffected because its container names match the default SAN prefix.

## Changes

* `--control-san-prefix n8n-sandbox-runner` so the SAN is `n8n-sandbox-runner-1`, matching the HTTP base URL host, the gRPC advertise host, and the network alias.
* Dropped `SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_SERVER_NAME`. gRPC now verifies against the dial host, same as HTTP.
* `compose.macos.yaml`: replaced unpublished image names (`n8nio/n8n-sandbox-api`, `-runner`, `n8nio/n8n-sandbox`) with the published `n8nio/n8n-sandbox-service-{api,runner-dind,sandbox}`.
* `AGENTS.md`: note that the example compose files are not CI-tested.
* Added ephemeral sandboxes to the smoke script.

Upgrade note: existing deployments must `rm -rf .tls` before `docker compose up -d`. `bootstrap-mtls.sh` skips generation when certs exist, and the old cert does not carry the new SAN.

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,6 +50,8 @@ Remember to update any relevant documentation in the docs/ folder if any of the 
 
 Document what the code does today, never what is planned.
 
+`docs/examples/compose.*.yaml` are not exercised by CI: when changing the API ↔ runner env or TLS contract, check for inconsistencies.
+
 ## Firecracker runner
 
 All Firecracker runner related files should contain `.ee` or be in a directory that contains `.ee` in the name, so the enterprise license applies.
--- a/docs/examples/compose.linux.yaml
+++ b/docs/examples/compose.linux.yaml
@@ -21,7 +21,7 @@ services:
         bootstrap-mtls.sh
         --out-dir /tls
         --api-san n8n-sandbox-api-local
-        --control-san-prefix n8n-sandbox-runner-local
+        --control-san-prefix n8n-sandbox-runner
         && chown -R sandbox-api:sandbox-api /tls/api
     environment:
       NUM_RUNNERS: ${NUM_RUNNERS:-1}
@@ -49,7 +49,6 @@ services:
       SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_CA_FILE: /tls/ca.crt
       SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_CERT_FILE: /tls/control-grpc-api-client.crt
       SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_KEY_FILE: /tls/control-grpc-api-client.key
-      SANDBOX_API_RUNNER_CONTROL_GRPC_TLS_SERVER_NAME: n8n-sandbox-runner-local-1
     healthcheck:
       test: ["CMD", "wget", "-q", "-O", "/dev/null", "http://localhost:8080/healthz"]
       interval: 5s
--- a/docs/examples/compose.macos.yaml
+++ b/docs/examples/compose.macos.yaml
@@ -16,7 +16,7 @@
 
 services:
   tls-init:
-    image: n8nio/n8n-sandbox-api:latest
+    image: n8nio/n8n-sandbox-service-api:latest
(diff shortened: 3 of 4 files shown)


tomi (approved):
LGTM 🚀 Tested macos locally and works

PR #175, opened 2026-09-10
feat: Separate the metrics route to a different port

njibhu:
## Summary

Adds a new env var SANDBOX_API_METRICS_LISTEN_ADDR which allows to have the metrics route on a separate http server, isolating it from the api routes.

## Review / Merge checklist

- [x] Tests included. 

<a href="https://cubic.dev/pr/n8n-io/n8n-sandbox-service/pull/175?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


njibhu:
@cla-bot check

njibhu:
Thanks — most of these were real and are fixed in cec2077. Two I want to push back on rather than change.

**`config.go:189` — validating the metrics address when `SANDBOX_API_METRICS_ENABLED=false`**

This is deliberate, not a parse-order accident. A malformed value the operator explicitly typed should fail loudly rather than sit latent until someone flips the enable flag, which is the worst moment to discover a typo. The same reasoning covers the gRPC-port collision check: validating it unconditionally means you learn about it now, whereas skipping it while metrics are off turns enabling them later into a fresh crash loop on a deployment that looked healthy.

Worth noting the premise is arguable — `SANDBOX_API_POSTGRES_*` is conditionally validated on `SANDBOX_API_STORE`, so there is in-repo precedent for the other style. The difference I care about is that the postgres vars are inert when unselected, whereas a metrics address is a thing you will switch on.

`TestLoadAPIMetricsListenAddrLoadsWhenMetricsDisabled` pins the part that does matter: disabled + a *valid* address loads fine and only warns, so the chart can set the address and the ServiceMonitor flag independently without coupling them.

**`main.go:212` — the shared shutdown context**

Correctly spotted, but a private context does not fix it. If the API drain burns the full 30s, the kubelet is already at `terminationGracePeriodSeconds` and about to SIGKILL the process, so a second 30s window buys the final scrape nothing; in the normal case the API drains in well under a second and the metrics server inherits ~29s, which is plenty for a sub-second scrape. Giving it its own deadline would only push the worst case past the grace period we are already being killed at.

The real defect was cosmetic: a spurious `ERROR metrics graceful shutdown failed` on an already-spent deadline, and a comment that oversold what the ordering guarantees. Both fixed in cec2077 — the shutdown now ignores `context.DeadlineExceeded`, since that case is the kubelet, not a failure.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

internal/api/config/config.go
```diff
@@ -178,6 +186,27 @@ func LoadAPI() (*APIConfig, error) {
 		cfg.GRPCListenAddr = v
 	}
 
+	if v := strings.TrimSpace(os.Getenv("SANDBOX_API_METRICS_LISTEN_ADDR")); v != "" {
```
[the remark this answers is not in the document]
njibhu: Keeping this as-is — it is deliberate rather than an artefact of where the block sits.

A malformed value the operator explicitly typed should fail loudly rather than sit latent until someone flips the enable flag, which is the worst moment to discover a typo. Same for the gRPC-port collision: validating it unconditionally means you learn now, whereas skipping it while metrics are off turns *enabling* them later into a fresh crash loop on a deployment that looked healthy.

The premise is arguable, though — `SANDBOX_API_POSTGRES_*` is conditionally validated on `SANDBOX_API_STORE`, so there is in-repo precedent for the style you suggest. The distinction I care about is that the postgres vars are inert when unselected, whereas a metrics address is a thing you intend to switch on.

`TestLoadAPIMetricsListenAddrLoadsWhenMetricsDisabled` pins the part that does matter: disabled plus a *valid* address loads fine and only warns, so the chart can set the address and the ServiceMonitor flag independently without the two being coupled.

cmd/api/main.go
```diff
@@ -166,6 +207,12 @@ func main() {
 	if err := srv.Shutdown(ctx); err != nil {
 		slog.Error("graceful shutdown failed", "error", err)
 	}
+	// Metrics last: a draining pod's final scrape is the one worth keeping.
+	if metricsSrv != nil {
+		if err := metricsSrv.Shutdown(ctx); err != nil {
+			slog.Error("metrics graceful shutdown failed", "error", err)
+		}
+	}
```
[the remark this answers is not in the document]
njibhu: Correctly spotted, but a private context does not actually fix it, so I kept the shared deadline.

If the API drain burns the full 30s, the kubelet is already at `terminationGracePeriodSeconds` and about to SIGKILL the process — a second 30s window buys the final scrape nothing, it just pushes the worst case past the point we are being killed at. In the normal case the API drains in well under a second and the metrics server inherits ~29s, which is ample for a sub-second scrape.

The real defect here was cosmetic: a spurious `ERROR metrics graceful shutdown failed` on an already-spent deadline, plus a comment that oversold what the ordering guarantees. Both fixed in cec2077 — the shutdown now ignores `context.DeadlineExceeded`, since that case is the kubelet rather than a failure, and the comment says so.

charts/n8n-sandbox-service/values.yaml
```diff
@@ -36,6 +36,9 @@ networkPolicy:
     # Additional peers allowed to reach the runner registration gRPC port.
     # The in-cluster runner is allowed automatically.
     grpcIngressFrom: []
+    # Peers allowed to reach the dedicated metrics port, if
+    # api.config.metricsListenAddr sets one. Empty allows all sources.
+    metricsIngressFrom: []
```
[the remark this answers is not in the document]
njibhu: This one is a false positive — no bump is needed, and nothing is failing.

`release-helm-chart-validate.yml` compares the chart version on HEAD against `origin/<base_ref>`, not against the previous commit:

```
base_chart=$(git show "origin/${{ github.base_ref }}:charts/n8n-sandbox-service/Chart.yaml")
...
if [ "$base_version" = "$head_version" ]; then
  echo "Chart changes must bump charts/n8n-sandbox-service/Chart.yaml version."
```

`main` is at `0.6.5` and this branch is at `0.7.0`, so `base_version != head_version` and the check passes — as the green `validate` job on this PR shows. Bumping again per commit would also be wrong here: `0.7.0` has not been released, so every chart change in this PR belongs to that one unreleased version.

charts/n8n-sandbox-service/templates/_helpers.tpl
```diff
@@ -110,3 +110,19 @@ app.kubernetes.io/component: {{ .component }}
...
 {{- printf "%s:%s" .Values.runner.sandboxImage.repository (.Values.runner.sandboxImage.tag | default .Chart.AppVersion) }}
 {{- end }}
+
+{{/*
+The API's dedicated metrics port, or "" when /metrics shares the HTTP port.
+Every metrics port, scrape and policy block in the chart keys off this.
+*/}}
+{{- define "n8n-sandbox-service.apiMetricsPort" -}}
+{{- $addr := .Values.api.config.metricsListenAddr | default "" | toString | trim -}}
+{{- if $addr -}}
```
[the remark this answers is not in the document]
njibhu: Valid, and fixed in c1c6301 — though narrower than suggested on two points, both deliberate.

**The guard is conditional rather than an outright rejection.** Loopback is a legitimate bind when the scraping happens *inside* the pod: a metrics-proxy sidecar shares the network namespace and reaches `127.0.0.1:9100` fine. So the render fails only when a ServiceMonitor — something outside the pod — would be doing the scraping:

```
api.config.metricsListenAddr ("127.0.0.1:9100") binds loopback, which the API ServiceMonitor
cannot scrape: bind all interfaces (":9100") or set monitoring.serviceMonitor.api.enabled=false
```

Every loopback spelling is covered (`127.0.0.1`, `127.0.0.5`, `localhost`, `[::1]` after bracket stripping), and `render-tests.sh` asserts both branches: the failure, and that it still renders with `monitoring.serviceMonitor.api.enabled=false`.

**Explicit interface addresses are not rejected, because they already fail loudly.** Both shapes were checked against the binary:

```
SANDBOX_API_METRICS_LISTEN_ADDR=10.11.12.13:9100
  ERROR metrics listen  error="listen tcp 10.11.12.13:9100: bind: cannot assign requested address"
SANDBOX_API_METRICS_LISTEN_ADDR=10.111.222.333:9100
  ERROR metrics listen  error="listen tcp: lookup 10.111.222.333: no such host"
```

Both exit 1 at startup, before the API port is opened. Inside a pod the only assignable addresses are loopback and the pod IP, and the pod IP is not knowable at render time, so a hardcoded interface address crash-loops with an accurate message and needs no chart-side guard.

Loopback is the only case that fails *silently*: it binds, serves on 127.0.0.1 (verified 200), and is unreachable from any routable address, so the target sits at connection-refused with nothing in the API log to explain it. That asymmetry is why it gets a render-time guard and an unassignable IP does not.

tomi (commented):
Good stuff 👌 Couple comments

tomi (commented):
Good stuff 👌 Couple comments

cmd/api/main.go
```diff
@@ -110,6 +117,29 @@ func main() {
 		IdleTimeout:       120 * time.Second,
 	}
 
+	// Bound before any goroutine starts, so an occupied port fails at startup
+	// rather than racing the signal handler.
+	var (
+		metricsSrv *http.Server
+		metricsLis net.Listener
+	)
```
tomi: Could we move this into separate package instead of inlining here? E.g. `metrics_server.go`

internal/api/metrics_router.go
```diff
@@ -0,0 +1,22 @@
...
+// NewMetricsRouter returns the handler for the dedicated Prometheus listener,
+// used when SANDBOX_API_METRICS_LISTEN_ADDR names a port other than the API's.
+// Deliberately no auth, access log, CORS or HTTP metrics: there is nothing else
+// on the port to protect, and a scrape is not API traffic.
+//
+// The Enabled guard is load-bearing: a disabled recorder has a nil registry,
+// which promhttp panics on at construction.
+func NewMetricsRouter(rec *metrics.APIRecorder) http.Handler {
+	mux := http.NewServeMux()
+	if rec.Enabled() {
```
tomi: Is it expected to call this when metrics are disabled? If not, it would be clearer to document the invariant by panicking if enabled is false. Something like:

```go
if !rec.Enabled() {
  panic("invariant violated: metrics must be enabled when creating a router")
}
```

api
tomi, line 1: I guess this shouldn't be here?

charts/n8n-sandbox-service/templates/configmap.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
 data:
   SANDBOX_API_LISTEN_ADDR: {{ .Values.api.config.listenAddr | quote }}
   SANDBOX_API_GRPC_LISTEN_ADDR: {{ .Values.api.config.grpcListenAddr | quote }}
+  SANDBOX_API_METRICS_LISTEN_ADDR: {{ .Values.api.config.metricsListenAddr | default "" | quote }}
```
endor: We seem to be repeating the default in multiple places?

charts/n8n-sandbox-service/templates/_helpers.tpl
```diff
@@ -110,3 +110,23 @@ app.kubernetes.io/component: {{ .component }}
 {{- define "n8n-sandbox-service.sandboxImage" -}}
 {{- printf "%s:%s" .Values.runner.sandboxImage.repository (.Values.runner.sandboxImage.tag | default .Chart.AppVersion) }}
 {{- end }}
+
+{{/*
+The API's dedicated metrics port, or "" when /metrics shares the HTTP port.
+Every metrics port, scrape and policy block in the chart keys off this.
+*/}}
+{{- define "n8n-sandbox-service.apiMetricsPort" -}}
+{{- $addr := .Values.api.config.metricsListenAddr | default "" | toString | trim -}}
```
endor: Just an idea: what if we bumped the chart major version and introduced a port here as default, so it's safe by default?

cmd/api/main.go
njibhu: I think its own package might be overkill, so I left it in the main package but as a separate file. We can always make a new package later if need to be, but I don't really see what we'd want to add there.

charts/n8n-sandbox-service/templates/_helpers.tpl
njibhu: This is only doable in another PR though since the changes are bundled together with the code change and the chart contains the app version, but this changes are not yet released.
I could split the PR in two between app and chart, but we could also just make a follow-up PR to only switch to a different port after we released the new version of the app.

cmd/api/metrics_server.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/n8n-io/sandbox-service/internal/api/config"
+	"github.com/n8n-io/sandbox-service/internal/metrics"
+)
+
+// newMetricsServer builds the dedicated Prometheus listener, or nils when
+// /metrics belongs on the API listener or is switched off altogether.
+//
+// The listener is bound here rather than in the serving goroutine, so an
+// occupied port fails at startup instead of racing the signal handler.
+func newMetricsServer(cfg *config.APIConfig, rec *metrics.APIRecorder) (*http.Server, net.Listener, error) {
```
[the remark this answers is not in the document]
njibhu: Tests are in cmd/api/metrics_server_test.go

endor (approved):

